PR #3, opened 2020-06-22
Fix fonts import

Trepix:
Hi, I detected the current URL is not downloading all the font-weights that theme uses.

`<link href="https://fonts.googleapis.com/css?family=Muli" rel="stylesheet" />`

Currently, it only brings 400 font-weight. Also, I have taken the opportunity to update the URL endpoint.

Due to this change, some parts of the web look different but I think brand text looks bad. For this reason, I adjusted the size from 800 to 600. I preferred to not touch more because the rest is readable and is consistent with the style. 

Please if you think that something has to be changed let me know.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -8,7 +8,8 @@
 <link rel="stylesheet" href="/assets/css/navbar.css" />
 
 <!--=================== cdn ==============================-->
-<link href="https://fonts.googleapis.com/css?family=Muli" rel="stylesheet" />
+<!-- <link href="https://fonts.googleapis.com/css?family=Muli" rel="stylesheet" /> -->
+<link href="https://fonts.googleapis.com/css2?family=Muli:wght@300;400;500;600" rel="stylesheet">
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.11.2/css/all.min.css" />
 
 <!--================= fab-icon =========================-->
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -8,7 +8,6 @@
 <link rel="stylesheet" href="/assets/css/navbar.css" />
 
 <!--=================== cdn ==============================-->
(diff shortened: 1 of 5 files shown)


hossainemruz:
Hi @Trepix thanks for the PR. Looks good to me.

I think headers also has the same problem as the brand text. Can you update the headers font-weight? Probably 400 will be more appropriate.
https://github.com/hossainemruz/toha/blob/632aa48c7116729188b9114561a51b375d093fca/static/assets/css/style.css#L26

Trepix:
Hi, I can do it easily. I'm just worried to break bootstrap consistency forcing all headers to the same weight. 
If you don't want to follow (or want to change) bootstrap style, please confirm it and I'm gonna apply the change.

hossainemruz:
The ultimate goal is to give the viewer a pleasant look and feel. So, if breaking the bootstrap consistency result better readability then let's be it. Try changing the weight. If you think, it looks better than the bootstrap's default look then change it. Maybe we can use different weight for different headers.

Trepix:
Are you concerned about a specific section? 
For me the result it's ok but if you can point me specific sections I can pay more attention.
If you don't have a specific concern I'm gonna screenshot both options and compare it

hossainemruz:
Don't you think that there is too much weight in the card header?

image

Also, headers in the reading page is too bold compared paragraph text.

image

If you think it doesn't look bad then let's go with the defaults.

Trepix:
I agree that the weight in these cases could be adjusted. But maybe it's better to remove the custom rule instead of forcing to apply 400 to all the headers.
I'm gonna work on it tacking into consideration the pages that you showed me

Trepix:
Hi @hossainemruz, I removed most of the font-weight rules. Bootstrap by default has 500, only one level more than the default level (400). 
I was comparing the section that you showed me (400 vs 500) and the difference is minimum. 
Are you ok with the result of the last changes? Do you think that something else is required?

hossainemruz:
Thank you @Trepix . Really appreciated your help.

PR #6, opened 2020-06-24
If no author specified in blog post, it should default to site author

jbleduigou:
Hi,

I have experienced a weird behaviour if I don't specify an author in the blog post.

For instance for the following blogpost:
```
---
title: "Test"
date: 2020-06-23T20:12:30+02:00
---
```
I would get John Doe as the author.

I think the helpers should look for the site author.


hossainemruz:
I was thinking the same. Just didn't get time to change. More refactoring is coming.

PR #9, opened 2020-06-25
Allow multiple position in one company experience

Trepix:
Hi, first of all, if this is not the appropriate channel to discuss it feel free to close it and show me where's the better place.

I'm working to have multiple experiences in the same company following the Linkedin approach. I think this might break the theme's consistency. I marked the most important changes (in my opinion)
Linkedin approach

If you are interested in this feature I want to discuss what's your idea to don't generate waste working on something that will have to be adjusted. If not, please, let me know to work on the feature without being worried to break the theme.


hossainemruz:
I think this is a reasonable feature. We have to discuss how the final design should look like. I will try to do a mock-up and get you back as soon as possible.

hossainemruz:
@Trepix  which pair do you prefer? (1,2) or (2,3)? Open Image in the new tab for full resolution.
Experience Timeline Mockup

For pair (1,2), we have to handle single and multi-position differently. For pair (2,3), both of them can be handle by same code.

In either case, `experiences.yaml` format should be as below

```yaml
experiences:
- company:
    name: Example Co.
    url: "https://www.example.com"
    location: Dhaka, Bangladesh
    overview: Example Co. is a widely recognized company for cloud-native development. It builds tools for Kubernetes.
  positions:
  - designation: Software Engineer
    start: Nov 2017
    end: Dec 2020
    responsibilities:
    - Design, develop and manage disaster recovery tool [Xtool](https://www.example.com) that backup Kubernetes volumes, 
      databases, and cluster's resource definition.
    - My another responsibility.
    - My more responsibilities.
  - designation: Junior Software Engineer
    start: March 2016
    end: May 2017
    responsibilities:
    - Write lots of example codes.
    - Read lots of examples.
    - See lots of example videos.
```

Trepix:
Hello,

I prefer de (1, 2) approach and I have some ideas for the multi-project view. I think some of them may fit.
When I show you the Linkedin approach was because I think that is a good way to show the information highlighting the most important part for each case.

Here are my proposals:

1. Follow the Linkedin approach of the experience headings section. They emphasize the designation in the jobs with only one project/designation, and in case you had multiple projects they put on top the company making it easier to read and 
avoiding to repeat information. 

2. The second is to eliminate the word _Responsabilities_ for the multi-project layout. I think that in cases that you were in more than 2 projects this only introduces noise. 

This is my first approach trying to take care of readability. Maybe the indent has to be improved if it's not easy to read, feel free to change it. I'm also trying to improve some paddings on mobile.

I attach desktop and mobile results. 

Desktop
image

Mobile
image

hossainemruz:
Do you prefer giving more emphasis on the `Role` for single position company like LinkedIn?

hossainemruz:
Then, go with the following design. Follow the `experiences.yaml` format as specified before.
Experience Timeline Mockup (1)

Trepix:
Hi,

> Do you prefer giving more emphasis on the `Role` for single position company like LinkedIn?

yeah I prefer to emphasize **what** i did. Put the company on top is a side-effect of avoid repeatition and make easy to organize the info.

Trepix:
Hello @hossainemruz,
I've added the changes to reproduce the schema you show in the attached snapshot. I've not updated the CSS because I think that is not necessary and could break the UX experience.

I've "fixed the bug" if the company's URL is not provided not add the link to `/`. 
Please, let me know if something has to be fixed/updated.

hossainemruz:
@Trepix I have pushed some changes. Added some comments, removed some extra codes, fixed CSS, and updated exampleSite.

PR #14, opened 2020-06-28
Make home page sections dynamic and configurable

hossainemruz:
**Why?**
- Give the control to the user over which section he want to show.
- Give the control to the user over the order of the sections.
- Give the control to the user over which menu he want to show on the top bar.
- Make it easy to add a new section.
- Don't show any unnecessary section when the user hasn't configured it yet.

**Warning: This PR will break the existing sites. See the comments for new configurations format.**

cc: @jbleduigou @Trepix @blueyetisoftware

PR #16, opened 2020-06-28
Allow section titles to be configured to match menu items

blueyetisoftware:


blueyetisoftware:
Looks like you beat me to it with #14 . I like that PR much better

blueyetisoftware:
I am reopening this one since I realized the names are not configurable on the section from #14 . This may need modification to be compatible with your other changes, but I am basically looking for a way to change the names and url for the different sections.

hossainemruz:
Perhaps, you can do something like,
```yaml
section:
  name: Portfolio
  enable: true
  weight: 6
  showOnNavbar: true
  template: projects.html
```
Then, on the `index.html` page,
```html
{{ if .section.template }}
  {{- partial .section.template . -}}
{{ else }}
  {{- partial (printf "%s.html" (replace (lower .section.name) " " "-")) . -}}
{{ end }}
```

blueyetisoftware:
PR replaced by #20

PR #20, opened 2020-06-30
Configure section title and menu urls in section yaml files

blueyetisoftware:


blueyetisoftware:
Each section added an additional div with a dynamic ID that matches the menu. This allows the urls to work while keep the original section ID to allow the stylesheets to match up. So each section ends up with 2 IDs (1 matching the stylesheet for the template, the other matching the dynamic name in the yaml file)

blueyetisoftware:
I recommend viewing the diff with whitespace changes hidden. Most of the changed lines are for additional tabs.

blueyetisoftware:
I wasn't sure if `-template` or `-section` was better for the section IDs. I kind of like section better. What do you think @hossainemruz ?

hossainemruz:
I think `section` seems better.

blueyetisoftware:
@hossainemruz I made the change to `-section`. Did the rest seem fine?

hossainemruz:
Looks good so far. I will test the PR against `hugoBasicExample` and `hugoThemes`  whenever get some time.

hossainemruz:
@blueyetisoftware I have updated the PR. Used `*-section` class instead of adding a `div`.  The previous approach had the following problems:
- It was breaking Responsiveness. Most responsive CSS was based on the top level `div`.
- It was possible to have same `id` for two different `div`. For example, if the user gave title `Project Section`, then the outer and inner div would have same `id`.

Please try the new changes and confirm if works for custom section name.

Do you have anything else to change? I am going to merge it after your confirmation.

blueyetisoftware:
@hossainemruz The updates work for me. Looks good. Nice changes

PR #2, opened 2020-06-21
Simplify code removing duplicated templates

Trepix:
If you don't have any intention or plan to have a different appearance for the left and right lines of the experience section, this could help to readability and maintainability.

Thanks for the work done with this nice theme! 😄

PR #22, opened 2020-06-30
Make soft skill's color configurable + add 'sky' and 'orange' colors

blueyetisoftware:


hossainemruz:
For Sky Color, which one do you prefer?

First one is the current one with `#01e0fd` hexcode.
image
In the second one I have made it little darker with `#00c9e3` hexcode.
image

ChapChap:
II think #2 is better looking `#00c9e3` among the other colors

hossainemruz:
@ChapChap  What if we make this color code configurable? :thinking:  Instead of color name, he will provide color code. User can provide whatever color he want.

ChapChap:
I think the best is to accept both options. The input is a string, if it is a color code, then the border-color is set to its value, else, seach its value among, predefined colors.
Most users (me included) struggle choosing the right color set... So for the convenience, its better to choose form predefined ones

hossainemruz:
Ok. Sound good to me.

hossainemruz:
@blueyetisoftware and @ChapChap I have made the colors configurable. Now, user can use either the pre-defined colors or he can provide his own color code. What do you think?

blueyetisoftware:
Great changes. I have been modifying the theme for my personal use by overriding the css. Being able to set the colors from the yaml is a nice feature.

PR #24, opened 2020-06-30
Allow section titles to be hidden inline

blueyetisoftware:


blueyetisoftware:
@hossainemruz I wanted to be able to hide the title for certain sections. What is the best way to have them enabled by default so the user doesn't need to set `showTitle`?

hossainemruz:
Can you please explain the use-case?

blueyetisoftware:
I wanted to show some of the sections without title as they are somewhat obvious. For example, the projects section is pretty clear as is the achievements section at the bottom. The achievements section is very graphical and interactive. The title didn't always fit with the images. It was just a styling preference for some cases.

hossainemruz:
Ok. Then, make showing title default behavior. Use `hideTitle`  instead of `showTitle`.

blueyetisoftware:
I was trying to match up with `showOnNavbar`. I am running into something similar with title showing up in the footer. I would like to use `showOnNavbar`, `showOnSection` and `showOnFooter`. What do you think?

hossainemruz:
>I would like to use `showOnNavbar`, `showOnSection` and `showOnFooter`.
I don't think we should add new variable for footer. If it was not shown on the navbar, then don't show in the footer too.

blueyetisoftware:
I have updated the PR per the conversation above. Should be ready

hossainemruz:
@blueyetisoftware Looks good so far. I will merge after some testing.

PR #26, opened 2020-07-01
Find first section for chevron navigation

blueyetisoftware:


layouts/partials/home.html
```diff
@@ -18,7 +18,8 @@ <h1 class="greeting">Hi, I am {{ if .Site.Params.author.name }}{{ .Site.Params.a
       <li>{{ . }}</li>
       {{ end }}
     </ul>
```
blueyetisoftware, line 20: I wasn't sure how to sort by `weight` and also filter for only `enable=true`. I tried something similar to this using `where` but wasn't able to get it to work.
https://discourse.gohugo.io/t/solved-sort-and-where-in-a-range-site-data/8225/2

blueyetisoftware:
I wasn't sure how to sort by `weight` and also filter for only `enable=true`. I tried something similar to this using `where` but wasn't able to get it to work.
https://discourse.gohugo.io/t/solved-sort-and-where-in-a-range-site-data/8225/2

layouts/partials/home.html
```diff
@@ -18,7 +18,8 @@ <h1 class="greeting">Hi, I am {{ if .Site.Params.author.name }}{{ .Site.Params.a
...
-    <!-- @todo this section should go to the first section, not necessarily about -->
-    <a href="#about"><i class="arrow bounce fa fa-chevron-down"></i></a>
+    {{ range first 1 (sort .Site.Data.sections "section.weight") }}
```
hossainemruz, line 21: You need to check if the `.Site.Data.sections` exist or not.

```html
{{ if .Site.Data.sections }}
 {{ range first 1 (sort .Site.Data.sections "section.weight") }}
    <a href="#{{ replace (lower .section.name) " " "-" }}"><i class="arrow bounce fa fa-chevron-down"></i></a>
 {{ end }}
{{ else }}
   <a href="#"><i class="arrow bounce fa fa-chevron-down"></i></a>
{{ end }}
```

blueyetisoftware: Updated with a check. If the sections don't exist, the chevron is hidden.

hossainemruz:
@blueyetisoftware I have pushed a change. Previously, the disabled sections was not handled properly. For example, if `About` section is disabled using `section.enable: false` it will be still present in the sorted map. Hence, the chevron was pointing to the about section.

PR #30, opened 2020-07-05
SubSections Anchors

VeskeD:
- Added anchors to sub-sections, to change cursor to hand

- To-Do - open this section(anchor) on link shared(openned)


VeskeD:
SubSections in /posts page is not displaying "hand" on hover cursor to it

VeskeD:
if somebody can do openning of exact anchor when section openned by link,

example:
https://toha.test.com/posts/#example-section

PR #31, opened 2020-07-06
Side BaseURL removed from single.html

VeskeD:
BaseUrl in single.html file been removed:

when background url is here with BaseURL, 
image is not openning when you accessing your page through some other url.

_example:_
when accessing site instead of: http:/toha.example.com/posts/some-post
you trying to access with: http://127.0.0.1/posts/some-post
or any other domain name

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -13,7 +13,7 @@
 {{ define "content" }}
 <div class="container p-0 read-area">
   <!--Hero Area-->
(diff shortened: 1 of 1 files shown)


VeskeD:
i'm a bit wrong here, it will be loaded from Damain if it's running,
but in case you are testing in local, it still will be loaded from Domain where file is located, but not from local.
and if site is not available through your domain name, it will not be loaded at all.

so this change is closing this little gap.

PR #32, opened 2020-07-07
Allow theme configuration to partially override css

phinatic:


hossainemruz:
@phinatic I think it will be better to make `style.css` configurable. Rename `style.css` into `main.css`. Then, rename `custom.css` into `style.css`.  Also, don't just import the css into a single page. Instead add it into `partials/header.html` All page use this header. So, they will get the changes.

phinatic:
I made the requested changes 👍

Issue #40, opened 2020-07-15
Footer does not route properly from posts page


al-haras:
Hello,

I noticed that when working with this theme, if you go to the posts page and click any of the navigation on the footer that it is not working correctly. I experienced this both using a local build and it exists on the example website as well.

From the posts page if you want to click on experiences, it will route to `https://toha.netlify.app/posts/#experiences` rather than (what I assume is expected) https://toha.netlify.app/#experiences.

![Screenshot_20200714_205234](https://user-images.githubusercontent.com/37927715/87501483-0b553300-c614-11ea-9a3f-646e334d9591.png)

I think the fix may just be a slight adjustment to your footer.html. If you agree that this is a bug, If I find some spare time to take a look and find and a good solution would you want me to create a pull request or is this something you'd prefer to look at and fix yourself?

StefanTobler:
I ran into this issue on my page as well. Good issue.

al-haras:
I do think I found an easy fix if you want me to create a PR with the change so you can take a look.

PR #41, opened 2020-07-15
add / to change URL in Navigation in footer

al-haras:
Fixes: #40


al-haras:
Possible solution for issue: #40

PR #46, opened 2020-07-22
Make Skills and Projects cards linkable

VeskeD:
Been added links to Card Headers to be able to open Project/Skill just by clicking on CardHeader, think it's good to have it clickable

### DIFF
--- a/exampleSite/data/sections/skills.yaml
+++ b/exampleSite/data/sections/skills.yaml
@@ -13,10 +13,12 @@ skills:
 - name: Kubernetes
   icon: "images/skills/kubernetes.png"
   summary: "Capable of deploying, managing application on Kubernetes. Experienced in writing Kubernetes controllers for CRDs."
+  url: "https://kubernetes.io/"
 
 - name: Go Development
   icon: "images/skills/go.png"
   summary: "Using as the main language for professional development. Capable of writing scalable, testable, and maintainable program."
+  url: "https://golang.org/"
 
 - name: Cloud Computing
   icon: "images/skills/cloud.png"
@@ -25,10 +27,12 @@ skills:
 - name: Docker
   icon: "images/skills/docker.svg"
   summary: "Write most of the programs as dockerized container. Experienced with multi-stage, multi-arch build process."
+  url: "https://www.docker.com/"
 
 - name: Prometheus
   icon: "images/skills/prometheus.png"
   summary: "Capable of setup, configure Prometheus metrics. Experienced with PromQL, AlertManager. Also, experienced with writing metric exporters."
+  url: "https://prometheus.io/"
 
 - name: Linux
   icon: "images/skills/linux.png"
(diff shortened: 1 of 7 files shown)


hossainemruz:
Sounds good. I will review and test ASAP.

hossainemruz:
@VeskeD What if we make the entire card linkable just like the recent-posts cards?

hossainemruz:
Btw, your change is breaking the CSS.
image

VeskeD:
> @VeskeD What if we make the entire card linkable just like the recent-posts cards?
i've tried that, but then it's not able to select text inside of card content, anyhow, choice is yours.



> Btw, your changes are breaking the CSS.
> image
ooopsy, i've did it's a bit different way in my page, thinking that this way will work without any css changes required,
will check this out.

as mine page is still using old version of theme, and i'm not yet updated with recent changes.

hossainemruz:
>i've tried that, but then it's not able to select text inside of card content, anyhow, choice is yours.

Ok. Then, lets make the entire card linkable.

>as mine page is still using old version of theme, and i'm not yet updated with recent changes.

Please, pull the latest changes from master. 

Let me know if you need any help with the CSS. Btw, is it ok for me to push in this PR? Since you have created this PR from your master branch, now any change we make in this PR will affect your master.

VeskeD:
> is it ok for me to push in this PR?

of course, go ahead if you really require.

> Let me know if you need any help with the CSS

I'm trying to build now separate page on my windows machine, instead of linux server, to test it out, 
but looks like you will be able to detect issue first :)

VeskeD:
@hossainemruz i've updated it, check this out now, should be good

VeskeD:
@hossainemruz need a bit help with solving this css issue, as it's now cards is with different sizes for some reason and indent between each other in columns
image

hossainemruz:
Cards height are adjusted from  `home.js`. I will take a look tomorrow.

VeskeD:
ok sure!

VeskeD:
@hossainemruz 

*Note: values is from your `toha-example-site`, check this out*

-------

i've updated `adjustSkillCardsHeight` because previous implementation is not working because of new <a> tag is there for skill cards.

after i updated, i've checked this issue, and for some reason after `home.js` is completed, it's setting up min-height as maximum available value, that's true, but as i can see, as page is not yet fully loaded, the maximum height was equals to "143px", so, this 143px value was set as `min-height` to `cards` and cards is not with same height after all content been loaded, but i have to say, that it's not always, if you will try to update this function and refresh page couple of times, sometimes it's taking correct `167px` value for `min-height`.

so, what i did is i make it calling after page is load, i know it's probably not the best solution, but now, `min-height` is always set to correct value equals to `167px` and as result all cards with same height.

probably you will have better solution without waiting while page is loaded, but at least now, cards now with same height.

PR #67, opened 2020-08-18
Adding background_position option

llaniewski:
This PR adds the option to set `background-position` css tag on the home page background image. For example to set it to always align to center of the image.

**Usage:**
```yaml
params:
  # background image of the landing page
  background: "images/background.jpg"
  background-position: center
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,9 @@ params:
 
 You can just copy the content for `config.yaml` files from `theme/toha/exampleSite/config.yaml`.
 
+If you want to customize the any of the CSS styles on your site, create a
+`static/assets/css/style.css` file and add any custom CSS there.
+
 ## Usage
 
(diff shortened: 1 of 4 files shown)


hossainemruz:
@llaniewski I think `background-position: center` should be default CSS instead of a configurable option in the `config.yaml`.  Please, change it here: https://github.com/hossainemruz/toha/blob/5f48a5bacb8a13cecaef5b9d9fa7efe25d860f79/static/assets/css/sections/home.css#L9

Btw, users can overwrite the default css with a `style.css` file in their `/static/assets/css` directory.

llaniewski:
@hossainemruz I set the `background-position` to `center` in the `home.css` and deleted the option, as per your comment. I also added the comment about the custom CSS to the README, as I think it will be useful to know for more tech-savvy users.

PR #112, opened 2020-10-21
Add nav item data and aria attributes to collapse the nav bar in mobile

alex-bezek:
…le when an item is clicked
https://github.com/hossainemruz/toha/issues/102

Tested on desktop chrome. Recreated the issue and then updated the nav links following https://getbootstrap.com/docs/4.0/components/collapse/
added a couple of aria attributes they recommended as well


hossainemruz (approved):
Thank you @alex-bezek .

PR #114, opened 2020-10-21
Fix about section to pull author name correctly

alex-bezek:
Addresses https://github.com/hossainemruz/toha/issues/113

Existing issue has name not showing up in the about section image

With this update it seems to pull correctly
image

### DIFF
--- a/layouts/partials/sections/about.html
+++ b/layouts/partials/sections/about.html
@@ -1,13 +1,19 @@
+
 {{ $sectionID := replace (lower .section.name) " " "-"  }}
 {{ if .section.id }}
   {{ $sectionID = .section.id }}
+{{ end }}
+
+{{ $author:= site.Data.author }}
+{{ if (index site.Data site.Language.Lang).author }}
+  {{ $author = (index site.Data site.Language.Lang).author }}
 {{ end }}
 
 <div class="container anchor p-lg-5 about-section" id="{{ $sectionID }}">
   <div class="row pt-sm-2 pt-md-4 align-self-center">
     <!-- summary -->
     <div class="col-sm-6">
-    <h3 class="p-1">{{ site.Params.author.name }}</h3>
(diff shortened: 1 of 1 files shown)


layouts/partials/sections/about.html
```diff
@@ -3,11 +3,16 @@
   {{ $sectionID = .section.id }}
 {{ end }}
 
+{{ $author:= site.Data.author }}
```
alex-bezek, line 6: This bit seems to be copied around a bit, but i'm unclear when this value would be set. 
My understanding is that this map follows this file format
image
Is it a possible convention to have an author file not in a locale folder?

hossainemruz (approved):

layouts/partials/sections/about.html
hossainemruz: >Is it a possible convention to have an author file not in a locale folder?

This is done for multi-language support.

alex-bezek: yeah, sorry I mean I understand that `(index site.Data site.Language.Lang).author` accesses the particular `data/:locale/author.yml` file. But i'm not clear on when `site.Data.author` would ever be not-nil. My assumption is that 
`{{ $author:= site.Data.author }}` always initializes the `$author` variable to nil and then its overridden in the if block below. 

I guess i can't just do 
```go
{{ if (index site.Data site.Language.Lang).author }}
  {{ $author := (index site.Data site.Language.Lang).author }}
{{ end }}
```
because go's variable scoping means the `$author` variable isn't declared outside the if. Can you explain if `{{ $author:= site.Data.author }}` ever is set to not nil and if so what config it uses? I'd like to understand your config.yml, site.yml, and locale specific fallback logic a little better to help address some of the issues. Is the thought that someone could make a file like `data/author.yml` which would make `site.Data.author` use those values? 

Thanks!

PR #115, opened 2020-10-21
add ios specific css for fixing the background images

alex-bezek:
Fixes https://github.com/hossainemruz/toha/issues/70

Read more into the comments of https://stackoverflow.com/questions/24154666/background-size-cover-not-working-on-ios which linked to https://stackoverflow.com/questions/30102792/css-media-query-to-target-only-ios-devices/47818418#47818418

I made that change and branch deployed it to my site to test
https://github.com/alex-bezek/blog/pull/50
https://blog-git-test-ios-background-image-fix.alex-bezek.vercel.app/
image

Tested in safari and firefox desktop as well
image
image

If anyone has an android and an ipad those would be good to test. I'll do some more reading about best practices using this css attribute as well since right now i'm just kind of trusting in stack overflow :)

### DIFF
--- a/static/assets/css/sections/home.css
+++ b/static/assets/css/sections/home.css
@@ -18,6 +18,19 @@
   background-size: cover;
 }
 
+/*
+  Resolves https://github.com/hossainemruz/toha/issues/70
+
+  fixed attached images use the whole <body> size. On mobile this can get really
+  tall which blows your image out. Setting the attachment back to scroll allows
+  your cover image to stretch within its own container
+*/
+@supports (-webkit-touch-callout: none) {
+  .background {
+    background-attachment: scroll;
+  }
+}
+
 .content {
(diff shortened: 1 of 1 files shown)


static/assets/css/sections/home.css
```diff
@@ -18,6 +18,19 @@
   background-size: cover;
 }
 
+/*
+  Resolves https://github.com/hossainemruz/toha/issues/70
```
alex-bezek, line 22: I can remove this comment if you'd prefer to rely on git history for this type of thing. I just typically like to add comments around "magic css" which is in place for a unique/special problem

hossainemruz: I think keeping this comment will be helpful.

hossainemruz (approved):

PR #121, opened 2020-10-24
Fix the site meta description to pull from the localized site data

alex-bezek:
### Issue
https://github.com/hossainemruz/toha/issues/108

### Description

The site's description used to live in the `config.yml` file. However, in the recent localization updates it should be moved to the `./data/<locale>/site.yml` file so its properly localized. The index file however is still referencing the old config option. This changes it to pull from the new localized section while also falling back to the top level config.yml for passivity. 

### Test Evidence

Before:
image

After:
image

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -2,7 +2,14 @@
 <html lang="en">
   <head>
     <title>{{- .Site.Title -}}</title>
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Looks good to me

PR #138, opened 2020-10-31
Fix the nav item collapse and spy functionality

alex-bezek:
### Issue

Fixes https://github.com/hossainemruz/toha/issues/132

### Description

Fixes an issue I introduced here https://github.com/hossainemruz/toha/pull/112/files when trying to fix the mobile nav not collapsing on click. It seems the ScrollSpy uses the data-attributes for determining which item to underline when scrolled to and all of them being the same causes this issue.

Gooing with another solution I had previously found https://github.com/hossainemruz/toha/issues/102#issuecomment-711502431

### Test Evidence

evidence


layouts/partials/navigators/navbar.html
```diff
@@ -31,16 +31,7 @@
     <div class="collapse navbar-collapse" id="top-nav-items">
       <ul class="navbar-nav ml-auto">
         <li class="nav-item">
-          <a
-            class="nav-link"
-            href="#home"
-            data-toggle="collapse"
-            data-target=".navbar-collapse"
-            aria-controls="collapseExample"
-            role="button"
```
alex-bezek, line 40: Breaking down the example more https://getbootstrap.com/docs/4.0/components/navbar/ they didn't use these for the actual links, these were for the item that toggled a drop down vs we just have a drop down when in mobile, so i'm removing these

hossainemruz (approved):
Looks good.

Issue #140, opened 2020-10-31
Author name is not showing in the post


hossainemruz:
## Expected Behavior

![image](https://user-images.githubusercontent.com/12577390/97782126-f8ae3c00-1bb9-11eb-830f-709203fbe683.png)

## Current Behavior

![image](https://user-images.githubusercontent.com/12577390/97782111-dddbc780-1bb9-11eb-9556-f19edc48b9b4.png)

## Steps to Reproduce

1.
2.
3.
4.

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

hossainemruz:
Fixed in #141

PR #141, opened 2020-10-31
Fix the get author name partial to account for localized author names

alex-bezek:
### Issue
https://github.com/hugo-toha/toha/issues/140

### Description

The author name seemed to have the same issue as the avatar did where it wasn't taking into account the new language-specific configs seen [here](https://github.com/hugo-toha/toha/commit/d0c32c5d082c4fa6b09690e60ebf06a4c59713a2#diff-fb24974d85736fe816c6ff9a4909af0d3faf111ade7fffe5b83a3b99072bf2c9R1) but the author name in this case had no fallback. 

### Test Evidence

Before
image
image

After:
image
image

### DIFF
--- a/layouts/partials/helpers/get-author-name.html
+++ b/layouts/partials/helpers/get-author-name.html
@@ -1,7 +1,12 @@
+{{/*  Uses the top level site's config for a single author across all locales  */}}
 {{ $authorName:= site.Params.author.name }}
-{{ if site.Data.site.author}}
-    {{ $authorName = site.Data.site.author.name }}
-{{ end}}
+
+{{/*  Overrides with the locale specifc author if provided  */}}
+{{ if (index site.Data site.Language.Lang).author }}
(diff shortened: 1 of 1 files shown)


layouts/partials/helpers/get-author-name.html
```diff
@@ -1,7 +1,12 @@
+{{/*  Uses the top level site's config for a single author across all locales  */}}
 {{ $authorName:= site.Params.author.name }}
-{{ if site.Data.site.author}}
-    {{ $authorName = site.Data.site.author.name }}
-{{ end}}
```
alex-bezek, line 4: correct me if i'm wrong, but I don't think the site.Data.site is every populated since there are language-specific site.yml files correct? Or do I need this for passivity for an old style of configuration?

hossainemruz: Yes. You are right. I am wondering how I missed that.

hossainemruz (approved):
Looks good. Thank you @alex-bezek.

PR #160, opened 2020-11-08
Remove unnecessary uses of baseURL

jragingfury:
Remove unnecessary uses of baseURL that could block the site from functioning if the URL isn't working. This primarily should only impact when migrating domains from one hosting service to another, but could also cause issues when the baseURL domain isn't resolving for some reason.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -35,7 +35,7 @@
   <div class="content">
     <div class="container p-0 read-area">
       <!--Hero Area-->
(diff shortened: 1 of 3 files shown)


hossainemruz:
@jragingfury this does not in demo in [hugo theme site](https://themes.gohugo.io/).

image

hossainemruz:
You can test against [hugoThemes](https://github.com/gohugoio/hugoThemes) repo.
- Follow the instruction [here](https://github.com/gohugoio/hugoThemes#testing-a-theme-with-the-hugo-themes-website-build-script). Use your own fork repo.
- After running for first time, cancel the run and go to `_script/reviewTheme/toha` directory.
- Checkout to your `baseurl-usage` branch.
- Run the script again.

PR #170, opened 2020-11-25
background issue and translation

vincentDcmps:
### Description

hello I have done two little modification:
first:
when we use theme with an URL like that: localhost/welcome
the default background was in 404 but I don't meet issue with either logo so I remove strings.TrimSuffix "/" site.BaseURL like for other logo

second: 
I have add I18n in footer contact to have translation for these field

third:
I have modified post translation for french 'des postes' give in English 'some computer'


layouts/partials/sections/home.html
```diff
@@ -28,7 +28,7 @@
 <div class="container-fluid home" id="home">
   <div
     class="background container-fluid"
-    style="background-image: url('{{ strings.TrimSuffix "/" site.BaseURL }}{{ $backgroundImage | relURL }}');"
+    style="background-image: url('{{ $backgroundImage | relURL }}');"
```
hossainemruz, line 31: This doesn't work on https://themes.gohugo.io/. That's why I had to keep it this way.

layouts/partials/footer.html
```diff
@@ -43,7 +43,7 @@ <h5>{{ i18n "navigation" }}</h5>
         <h5>{{ i18n "contact_me" }}</h5>
         <ul>
           {{ range $key,$value:= $author.contactInfo }}
-          <li><span>{{ title $key }}: </span> <span>{{ $value }}</span></li>
+          <li><span>{{ i18n $key }}: </span> <span>{{ $value }}</span></li>
```
hossainemruz, line 46: We don't need i18n translation here. The author information will be on their own locale (i.e. `en`/`fr` folder). So, they will be already translated there.

hossainemruz (changes requested):
Thank you @vincentDcmps for your PR. Sorry for being late to review. I couldn't make some time to work on it. Btw, I have requested some changes. Can you please address them?

layouts/partials/sections/home.html
vincentDcmps: same with base URL set it's strange. how do you set BaseURL on my it's like that:
``baseURL : "https://www.ducamps.win/welcome"``

hossainemruz: @vincentDcmps I am afraid I didn't understand your comment properly. Can you explain little bit?  Did you mean it doesn't work with `baseURL : "https://www.ducamps.win/welcome"`?

vincentDcmps: With this baseurl it s work only with removing background-image: 
''url('{{ strings.TrimSuffix "/" site.BaseURL }}'' 
It s weird to put this only for background ans not for other logo

vincentDcmps: to reproduce but in config.yaml 
```baseURL: 'https://www.ducamps.win/welcome```
and launch test server with ```hugo server -b localhost/welcome -w```

hossainemruz:
Sorry @vincentDcmps. I can't merge this PR. Background image don't show up in [Hugo Theme site](https://themes.gohugo.io/) with your change.
image

vincentDcmps:
oh I don't understand why I have reverse effect with my config

Issue #171, opened 2020-11-26
Use Hugo's Image Processing


patmagauran:
The theme should use [Hugo's built in image processing](https://gohugo.io/content-management/image-processing/) throughout the project, both in the posts as well as the main page layout. I think this can be done through a global resource and the appropriate functions.

RotatingFans:
I did a test with the Avatar image on the home page and was able to get it to work, so it shouldn't be too hard to do. I will try to get a pull request together soon. However, it will require moving the images from the static folder to the assets folder. So we will need to update the docs as well.

hossainemruz:
>However, it will require moving the images from the static folder to the assets folder.

Probably this was the reason why I didn't use Hugo's native image processing. Is there any other possible issues that may cause by moving the images outside of the static folder? Obviously, it's going to be a breaking change.

RotatingFans:
I am not deeply familiar with Hugo, but overall the change seems pretty painless. No changes are needed to the configuration, just copy the images folder from /static to /assets. Obviously this is a breaking change, so it would be important to determine a way to override this. Perhaps a configure option: useImageProcessing, that will allow for the images to be pulled from the asset folder, which would be false be default when upgrading.

hossainemruz:
Done in #173

PR #173, opened 2020-11-26
Transition to Hugo Image Processing

patmagauran:
### Issue
#171 

### Description
This change will changeover all user-configurable images within the layout to utilize Hugo's Image Processing which allows for images to automatically be scaled and optimized, increasing page load performance.

It also implements a basic shortcode `{{<rimg src="Export-25.jpg" alt="A pretty image">}}` that will allow for the `srcset` attribute to be used, allowing compatible browsers to load the appropriately sized image in posts.

### Changes
- Images are loaded using the Hugo Image Processing functionality
- Images must be loaded from the `/assets` folder instead of the `/static` folder. **This is a breaking change**

### Mitigation of Breaking Changes
It would be possible to check for the value of a site-wide config that will dictate whether to use this way or the standard way. This could allow for a transition period, wherein the user has to make the switch. 

Changing is as simple as copying the files

Also possible would be to in case of failure to load image via hugo processing, fall back to old way and hope they are still in `/static` folder.

### Test Evidence
Copy images from `/static` to `/assets` and use this version of the theme. Images will be appropriately resized and or scaled.


RotatingFans:
Two more issues that need to be addressed(Based on further testing with my personal site): 
- Hugo Image Processing only supports jpg, png, tif, bmp, and gif. SVG is not supported. Therefore, we need to check for this before trying to resize
- If the supplied path is not valid, it errors out. It should display nothing instead, or the missing image logo.

hossainemruz:
@RotatingFans Sorry for being late. It may take few more days. I just couldn't make time to test it.

hossainemruz (approved):

dyingway:
with this change, i don't know how to set the hero jpg address,before the change,i set the hero like the link
`/assets/images/hero/winter-1142029_1920.jpg`

now , I can't get a good understanding of how the new code
`themes\toha\layouts\partials\helpers\get-hero.html`



works，please help how to set hero link,if the images as the link `/assets/images/hero/winter-1142029_1920.jpg`
thank you so much

hossainemruz:
Just put your hero image in the same folder as you post markdown. Rename your hero image as `hero.png`, or `hero.jpg` or `hero.svg`. You don't have to specify anything in the post-front matter.

dyingway:
i test but can't work...
i test Different catalogues。

`\content\Pro-bono`,   as `\content\Pro-bono\hero.jpg`

the md `\content\Pro-bono\about.md` not show the hero.

i **also test** set in `\content\Pro-bono\about\hero.jpg`

not change...


i aslo change the post-front matter:

hero: hero.jpg
or
del `hero:`

I don't know what I did wrong，thank you for your patience.

hossainemruz:
Can you point me to your repository?

hossainemruz:
You can follow the sample post structure as [this](https://github.com/hugo-toha/hugo-toha.github.io/tree/source/content/posts/introduction)

dyingway:
I know where I went wrong.
If I want to show hero, I have to create a folder for any file and name it `index.md` before hero work.
it work,when i change about.md to index.md and put it in`\content\Pro-bono\about\'

*****
I don't think this is a good change. Because it would mean that I would have to create a folder for each file, which might instead break the readability and edit management of the files, if there are a lot of md.

Wish there was a better way to manage the hero.and I think HERO is the background, not the content.thank you for your hard work：）

hossainemruz:
This structure is recommended by hugo. ref: https://gohugo.io/content-management/organization/

patmagauran:
I do see your concerns about the readability and management of files, but I think this style of content management actually makes it more manageable, especially for larger sites. It proactively prevents issues where you may add a resource to a page without having the dedicated folder for it. It breaks up each page on the content side cleanly into its own little folder. It prevents the clutter of having `hero-image-1234567.jpg` and so forth all in one folder. It is unfortunate that it restricts users to a predefined organization, but it is easier to do it from the start that way.

PR #197, opened 2020-12-26
explicitly set the default language (needed for English not being the…

alteholz:
In case you don't want to use English as the default language, you need to set a different language with "defaultContentLanguage" in the config file. Otherwise the index page still would be rendered with English.


hossainemruz (approved):
Thank you @alteholz

PR #203, opened 2020-12-29
Make list.html fully generic

hblunck:
I love the list template. However, I had some issues using it for other content archetypes other than `posts`. This code change makes list.html fully generic, there are no more built in references to `posts`.

Now, when adding a new archetype, all the user needs to do is add the name to the i18n documents.


hossainemruz (approved):
Thanks @hblunck

PR #210, opened 2021-01-05
Translate chinese

Chow-Ray:
### Description

Translate to Chinese, fix weight serial number（only fix English & Chinese）。


hossainemruz (approved):
Thank you @Chow-Ray

PR #232, opened 2021-02-16
Add i18n to experiences section

Roxedus:
### Description

Add some missing translations for the experiences section


hossainemruz (approved):
Thanks @Roxedus

PR #256, opened 2021-03-22
Fix missing vertical separator in navbar menu

### LINKED ISSUE #255: Menu separate does not appear if blog post is disabled.
## Expected Behavior
image

## Current Behavior
image

## Steps to Reproduce

1.
2.
3.
4.

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

chutified:
### Issue

Resolve #255

### Description

I've created a new variable `$customMenus` which has by default a value of `site.Params.customMenus`.
The problem causing the bug was that it didn't check the language-scope `customMenu` value.

But it does now :)

*The same value is being reused to iterate over those custom links and to display them, so I have replaced it with the variable.*

### Test Evidence

menu
menu-2
menu-3

### DIFF
--- a/layouts/partials/navigators/navbar.html
+++ b/layouts/partials/navigators/navbar.html
@@ -23,6 +23,11 @@
 {{ end }}
 {{ $invertedLogo = $invertedLogo.RelPermalink}}
 
+{{ $customMenus := site.Params.customMenus }}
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank you @chutified . Looks good.

PR #278, opened 2021-04-01
set url target link to blank

HenzelMoras:
### Description
url redirection fix

### Test Evidence
new tab

PR #288, opened 2021-04-03
Fix inconsistencies in target attributes

chutified:
### Description

Currently there are two ways how the code opens links in new tabs:

* `target="/"`
* `target="_blank"` (in #278)

Both attributes work well but there shouldn't be two ways of doing one thing.

I didn't manage to find how exactly the first one (`target="/"`) works so I decided to go with the second
solution (`target="_blank"`). Also according to these resources this is the "correct" way to handle these situations:

https://www.w3schools.com/tags/att_a_target.asp (`"/"` is not even an option)
https://css-tricks.com/all-about-mailto-links/#open-in-new-tab-sometimes-does-matter (`"_blank"` is used in the same context)

If there is any particular reason to use `"/"` please let me know.

### DIFF
--- a/layouts/partials/sections/about.html
+++ b/layouts/partials/sections/about.html
@@ -29,11 +29,11 @@
           {{ range .socialLinks }}
           <li>
             {{ if eq .name "Email" }}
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank  you @chutified

PR #289, opened 2021-04-04
add video shortcode

davidgs:
A video shortcode so you can include `{{< video "/path/to/video.mp4in pages. " >}}`


hossainemruz:
@davidgs Thank you for the PR. I will review it soon. We probably need to change how scripts has been handled in this PR.

hossainemruz:
We should also explore which will better choice.
- https://github.com/sampotts/plyr
- https://github.com/clappr/clappr

davidgs:
Yeah, I've been looking at how scripts and stye sheets are included overall because they really should be shar-256 verified so that we don't get 
`[Error] Refused to apply a stylesheet because its hash, its nonce, or 'unsafe-inline' does not appear in the style-src directive of the Content Security Policy.` 
errors when `Header always set Content-Security-Policy` is enforced the scripts and style sheets get blocked.

Nirzak:
> Yeah, I've been looking at how scripts and stye sheets are included overall because they really should be shar-256 verified so that we don't get
> `[Error] Refused to apply a stylesheet because its hash, its nonce, or 'unsafe-inline' does not appear in the style-src directive of the Content Security Policy.`
> errors when `Header always set Content-Security-Policy` is enforced the scripts and style sheets get blocked.

Yeah but that's only for particular sites as Github Pages doesn't support custom headers. Butw I will like you guys to handle the background image script from external script instead of using inline script. It will help us not to enforce CSP with unsafe-inline

hossainemruz:
@davidgs I have moved to [sampotts/plyr](https://github.com/sampotts/plyr) library for video player. It is lightweight and handles responsiveness well.

hossainemruz (approved):

hossainemruz:
Pull the latest `main` branch and let me know if it works.
You have to user the shortcode as bellow,
```
{{< video src="/videos/sample.mp4" >}}
```

PR #290, opened 2021-04-04
Fix redirections to empty tabs

chutified:
### Problem

Whenever href (url) attribute was omitted or had no value
the click action would do nothing. With no URLs defined and
`target="_blank"` attributes, which were recently added,
anchor tags now redirect users to new blank tabs (side effect).

### Solution

This PR solves the issues by adding conditions
before applying `target="_blank"` or in some cases by
ommiting entire entities (e.g. the view_certificate button).


Nirzak (approved):

hossainemruz (approved):
Thank you @chutified . Looks good.

PR #292, opened 2021-04-05
Refactor target values of redirections to external pages

chutified:
## Proposal

I believe we have really improved UX by opening **section external links** in new tabs (#278). On the other hand, I don't think we should stop there. I would do the same for **all links** that redirect to external sites, which typically have no backlinks to our website. Therefore, I think it would be nice to open all these links in new browser tabs as well.

### Changes

This commit adds three `target="_blank"` attributes to anchor tags:

* in the footer:
  * gh repo of the toha theme
  * hugo official website
* _"improve this page"_ button.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -59,7 +59,7 @@
         <!--- Improve this page button --->
         {{ if site.Params.GitRepo }}
           <div class="btn-improve-page">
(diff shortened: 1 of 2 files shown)


hossainemruz (approved):
Thank you @chutified

PR #296, opened 2021-04-19
Accessibility and SEO improvements

jcabak:
- Image elements have `[alt]` attributes
- All links have `title` attribute and are accessible to everyone
- Links to cross-origin destinations are safe

### DIFF
--- a/layouts/partials/sections/home.html
+++ b/layouts/partials/sections/home.html
@@ -127,7 +127,7 @@
         {{ if .section.id }}
           {{ $sectionID = .section.id }}
         {{ end }}
-        <a href="#{{ $sectionID }}"><i class="arrow bounce fa fa-chevron-down"></i></a>
+        <a href="#{{ $sectionID }}" aria-label="{{ i18n "read" }} {{ i18n "more" }} - {{ $name }}"><i class="arrow bounce fa fa-chevron-down"></i></a>
       {{ end }}
     {{ end }}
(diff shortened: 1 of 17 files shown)


layouts/partials/sections/home.html
```diff
@@ -127,7 +127,7 @@ <h1 class="greeting"> {{ $author.greeting }} {{ $name }}</h1>
         {{ if .section.id }}
           {{ $sectionID = .section.id }}
         {{ end }}
-        <a href="#{{ $sectionID }}"><i class="arrow bounce fa fa-chevron-down"></i></a>
+        <a href="#{{ $sectionID }}" aria-label="{{ i18n "read" }} {{ i18n "more" }} - {{ $name }}"><i class="arrow bounce fa fa-chevron-down"></i></a>
```
Nirzak, line 130: Can't we just write it in a single braces like ```aria-label="{{ i18n "read_more"}} - {{ $name}}"``` instead of ``` {{ i18n "read"}} {i18n "more"}} ```

jcabak: We can't. There is not such a thing like `"read_more"` phrase in i18n files so we need join two of them. I tried, it's not working.

Nirzak (approved):

hossainemruz (approved):
Thank you @jcabak . Looks good to me.

Issue #224, opened 2021-01-21
Modify header and footer as if they were just another section


jpmvferreira:
Being able modify the header and the footer as if they were another section would make it easier to fully customize the website.

For example I would like to keep things as clean as possible and given that I have no intention to translate to other languages having the language on the top right can be misleading. Commenting it out is an option but having the intention to not having the logo on the header and centering the rest of the menu is not an option.

The same thing applies for the footer, allowing the user to customize a set of parameters.

Kaligule:
One great feature here would be to allow to completly disable header and footer. Personally I would like to get rid of the footer entirely for my site.

hossainemruz:
Ok, we can do this. Just need some free time.

Kaligule:
So this might be a stupid question, but is the flag-problem maybe solved already?

The flags are inserted into the navigation bar [here](https://github.com/hugo-toha/toha/blob/v2.2.0/layouts/partials/navigators/navbar.html#L106):

```html
{{ if .IsTranslated }}
  {{ partial "navigators/lang-selector.html" . }}
{{ end }}
```

`IsTranslated` is a variable set by hugo itself.
The [hugo documentation](https://gohugo.io/variables/page/) tells us about it:

```
.IsTranslated
    true if there are translations to display.
```

And indeed, on my site there are no flags to see:

![Screenshot_2021-04-28_13:56:46](https://user-images.githubusercontent.com/3586246/116399820-9f9e2880-a829-11eb-8613-2fbbc4dd0af1.png)

So how do we know if there are translations?
I am not sure, but I guess removing the "langues" section from your site config might do the trick.
@jpmvferreira  Could you give this a try?

hossainemruz:
Made footer configurable in #303 . Thanks to @Kaligule . You can now enable or disable a particular section of the footer. You can even provide a custom footer template.
Example: https://github.com/hugo-toha/hugo-toha.github.io/blob/5fcc45e97514e1c206a91ab1291b0da9d65da758/config.yaml#L69

Kaligule:
@hossainemruz Thank you. I think the navbar (what is probably what @jpmvferreira meant with "header") is not yet configurable, right? So we should either reopen this issue or open a new one.

hossainemruz:
The Navbar is already configurable. User can show/hide a particular section in the navbar

PR #303, opened 2021-04-28
Allow footer configuration

Kaligule:
Introduced some new configuration options. These are the default values:

```yaml
params:
  footer:
    enable: true
    navigation:
      enable: true
    contactMe:
      enable: true
    credentials:
      enable: true
  newsletter: # not new, but similar to new options
    enable: false
```

Where should the documentation go?

Please note that I am technically a web developer, but not familiar with hugo in the slightest. I am very happy about constructive feedback.

Btw it might be easier to review the commits one by one because the indentation makes the diff look more crazy than it is.

### Issue

https://github.com/hugo-toha/toha/issues/224

Doesn't solve that issue completly as it is about the footer _and the navbar_. The navbar was a bit over my head at this moment.

### Description

Allows to disable/enable parts of the footer (or all of the footer).

#### Footer disabled

Screenshot_2021-04-28_12:59:29

#### Navigation disabled

Screenshot_2021-04-28_12:58:53

#### Footer enabled, but all parts of it disabled

Screenshot_2021-04-28_12:58:14

#### Copyright notice disabled
Screenshot_2021-04-28_13:01:28


hossainemruz:
@Kaligule  Thank you for the PR. I think footer should be just another section so that user can use custom template for it.

Kaligule:
@hossainemruz I don't think I can help there. Do I understand you correctly that you do not want to merge this PR but work on your own solution? (I this case I will have to point my site to my own fork until you succeded.)

hossainemruz (approved):
Thanks @Kaligule . I have made little adjustment. Now, user can provide a template for the footer.

Kaligule:
Cool, thanks :)

PR #311, opened 2021-05-14
In case the old config already deactivated the newletter, it must be …

alteholz:
…also deactivated in the compatibility code.


hossainemruz (approved):
Thank you @alteholz

PR #326, opened 2021-05-24
Fix German translation + add Dutch translations

serverfriendsorg:
### Issue
This PR fix the german translation as it was not complete and also contained some dutch strings.

### Description
The german translation was refactored and the dutch translation separated into a own file.


hossainemruz (approved):
Thank you @serverfriendsorg

PR #330, opened 2021-05-25
Add configuration to number of recent posts.

donfiguerres:
### Description

Add configuration to the number of recent posts.


donfiguerres:
I'm going to close this because I just realize it will break sites that do not have the config yet. I'll create a new one that sets a default number if the config value is not found.

PR #336, opened 2021-06-01
Disable logo by default + add logo only if the user provide it

Kaligule:
### Issue
There is no related issue.

### Description

Logos are not shown by default. (I figured this is a sensible default because most people don't have a personal logo. Especially not in different light/dark version. But I would be happy to discuss this point.

Logos can be enabled again with (toml)

```toml
[params.logo]
enable = true
```

or (yaml)

```yaml
params.logo.enable: true
```
### Test Evidence

no_logo

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -1,3 +1,5 @@
+{{ $logoEnabled   := site.Params.logo.enable  | default false }}
+
 {{/* default favicon */}}
 {{ $favicon := "/images/favicon.png" }}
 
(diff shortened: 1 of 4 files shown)


hossainemruz:
Let's keep logo enabled by default. We can have a parameter to disable it. Or, if the logos are not provided, then disable it  instead of defaulting to theme logo.

Kaligule:
Seems reasonable. I will see when I can work on this.

hossainemruz (approved):
Thank you @Kaligule . I have made some changes to the PR. Now, the logo will be added only if  the user provide it. It won't default to theme logo anymore.

hossainemruz (approved):

PR #341, opened 2021-06-03
Fix `ressource.Fit` not resizing the image properly

### LINKED ISSUE #340: Difference ressource.Resize vs ressource.Fit
Hi,

First of all, congratulation for your wonderful theme and documentation. I tried various themes and it is by far the most complete (and visualy nice) theme I played with

After going deeper into Hugo code, I am wondering what is the difference between ressource.Resize and ressource.Fit
https://gohugo.io/content-management/image-processing/

For example, for the logos of the skill section, I realize that despite the use of ressource.Fit in your code (24x24px), the full  original picture is downloaded when loading the page.

Wouldn't it be more efficient to use ressource.Resize. Moreover, could you explain to me what is actually doing ressource.Fit.
I searched a bit for detailled explanation but wasn't successfull.

Thanks
Valérian

vlebert:
Fixes #340

### DIFF
--- a/layouts/partials/cards/project.html
+++ b/layouts/partials/cards/project.html
@@ -11,7 +11,7 @@
 
             {{ $logoImage:= resources.Get .logo}}
             {{ if $logoImage }}
-            {{ $logoImage := $logoImage.Fit "24x24" }}
+            {{ $logoImage = $logoImage.Fit "24x24" }}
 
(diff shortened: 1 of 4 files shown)


hossainemruz (approved):
Thank you @vlebert . Looks good to me.

PR #342, opened 2021-06-03
Skillcards height css instead js

vlebert:
### Description

Lighter than JS (CSS only), and gives me better result when the cards conains a bit more text. 
Expecially when you resize the window after page load.

What do you think about this approach?

### Test Evidence

Here is an example with these changes
https://vleb-dev.netlify.app/


hossainemruz (approved):
Thank you @vlebert. Looks good to me.

PR #347, opened 2021-06-05
Fix horizontal scrollbar

### LINKED ISSUE #304: Horizontal scroll bar on small devices
I noticed a vertical scroll bar on smaller screens (width: less than 618px). It is there for an excess white small stripe on the right side. I couldn't exactly find the source of the problem, but it seems to have something to do with the `skills` section because the moment I remove this section in the chrome inspection the horizontal scroll bar disappears and only reappears when the width of the page is shrinked below the min width of some divs.

## Steps to Reproduce
1. Visit https://hugo-toha.github.io/.
2. Open developer tools.
3. Reduce the width to less than 618px.

toniop99:
### Issue
Fixes #304 

### Description

Changes some global styles just for adjust the ` <h1> ` texts to the screen size.
Changed on the css of about and skill sections to fix some problems on small screens.
On about sections this problems was related to the **badges** and **sof-skills** grid.
Fixed it adjusting the size of this item while the screen is getting smaller and ending on the smallest screens with just one column.

### DIFF
--- a/static/css/layouts/main.css
+++ b/static/css/layouts/main.css
@@ -432,16 +432,29 @@ mark {
   code {
     padding: 0px;
   }
+
+  h1 {
+    font-size: 2.2rem;
+  }
 }
 
 /* iPhoneX, iPhone 6,7,8 */
 @media only screen and (max-width: 375px) {
+  h1 {
+    font-size: 2rem;
+  }
 }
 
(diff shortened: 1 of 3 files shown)


hossainemruz (approved):
Thank you @toniop99 . I can confirmed that this fixes the issue.

PR #350, opened 2021-06-09
Add greeting subtitle

vlebert:
For my website I'd like to add an optionnal subtitle.

See changes. Would you be interested to add the feature in the main project?

### DIFF
--- a/i18n/fr.toml
+++ b/i18n/fr.toml
@@ -3,7 +3,7 @@
 other = "Accueil"
 
 [posts]
-other = "Des postes"
+other = "Articles"
 
 [toc_heading]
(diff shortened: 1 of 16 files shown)


hossainemruz:
Hi @vlebert!  Thank you for the PR. Can you please share a screenshot so that we can see how it looks?

vlebert:
Hi @hossainemruz 

For the moment the result is very simple. I was thinking more style element could be added afterward in the custom `style.css`. I tried to set font to italic but the result was ugly ^^

I am open to artistic suggestion though :)

image

hossainemruz (approved):
Thanks @vlebert

PR #353, opened 2021-06-16
Support Google Analytics v4

### LINKED ISSUE #352: bug: Google Analytics 4 is not supported
I want to start by saying thank you - I love this theme and have found your docs fantastic!

Unfortunately, I'm not able to use **Google Analytics v4** with the toha theme. **Google Analytics 4** is currently the default for new accounts.

Looking at your layouts, I see the following code in both the `index.html` and the `baseof.html` files: 

```html
    {{ if site.GoogleAnalytics }}
        {{ template "_internal/google_analytics_async.html" . }}
    {{ end }}
```
From the [HUGO documentation](https://gohugo.io/templates/internal/#use-the-google-analytics-template), for Google Analytics v4:

> When using Google Analytics v4 use `_internal/google_analytics.html`.

The `_internal/google_analytics_async.html` which toha is using seems to only work with v3.

As a workaround, you can still get the old v3 IDs by following [this post](https://support.google.com/analytics/answer/10269537), but you can't leverage a lot of the stuff that is in v4...

Thanks again!

Dan

gilcrest:
Per [Issue #352 ](https://github.com/hugo-toha/toha/issues/352): in order to support Google Analytics v4, _internal/google_analytics.html should be used

Fixes #352

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -11,7 +11,7 @@
 
     <!-- ADD GOOGLE ANALYTICS IF ENABLED -->
     {{ if site.GoogleAnalytics }}
-      {{ template "_internal/google_analytics_async.html" . }}
+      {{ template "_internal/google_analytics.html" . }}
     {{ end }}
   </head>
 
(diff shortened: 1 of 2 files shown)


hossainemruz (approved):
Thanks @gilcrest

gilcrest:
@hossainemruz - again, thanks so much for the great work!

PR #369, opened 2021-07-03
Remove about section whitespace when there are no soft skills/badges

### LINKED ISSUE #368: About section summary should take up entire page width if no soft skills are given
Before filing an issue, please review the following:
* Existing Features: https://github.com/hossainemruz/toha#features ✅
* Project Roadmap: https://github.com/hossainemruz/toha#project-roadmap ✅

I've been building my personal site with Toha and really enjoying everything so far. However, I don't plan on including any soft skill "circles" in my About section. If I don't specify any `softSkills` in my `about.yaml` file, the template leaves a big chunk of whitespace to the right of the summary.

current_about_section_with_whitespace

I think it would look nicer if the summary took up the entire width of the section if no soft skills are specified like what's pictured below. 

about_section_without_whitespace

~~I am very new to Hugo templates but I believe the following addition to [line 14 of `about.html`](https://github.com/hugo-toha/toha/blob/4099f4f1e65299866e739b3be43b6d147ab16e61/layouts/partials/sections/about.html#L14) would implement this feature nicely.~~ 

The following solution is now outdated as I forgot to consider the presence of [badges](https://github.com/hugo-toha/toha/pull/307). Please see #369 for the updated code.  

proposed_solution(1)

In the future, it may also be nice to give the user an option to include a rounded profile picture to the right instead of soft skills. But that might be an idea for a different issue. 

Would be happy to make a PR with this change if all looks good. Thanks!

will-hou:
### Issue
Closes #368. The about section summary will now take up the entire section width when no soft skills/badges are specified in `about.yaml`

### Description

This PR adds a conditional statement that checks whether soft skills (indicated by `softSkills` in `about.yaml`) or badges (indicated by `badges`) have been specified by the user. If there are soft skills/badges, the original `col-sm-6` style is applied. However, when there are no soft skills, `col-sm-12` is applied instead — indicating that the summary should take up the full width of the section. 

### Test Evidence

I tested all changes locally using Firefox 89.0.2. Formatting when soft skills are specified, badges are specified, or both remain unchanged across both desktop and mobile versions of the site. When neither soft skills nor badges are specified, the about section summary takes up the entire width on both desktop and mobile versions of the site. All other formatting remains unchanged. 

aboutr_section_without_whitespace

The summary also displays correctly on Chrome and Edge.

### DIFF
--- a/layouts/partials/sections/about.html
+++ b/layouts/partials/sections/about.html
@@ -11,8 +11,8 @@
 <div class="container anchor p-lg-5 about-section" id="{{ $sectionID }}">
   <div class="row pt-sm-2 pt-md-4 align-self-center">
     <!-- summary -->
-    <!-- takes up full section width if no soft skills are specified -->
-    {{ if .softSkills }} <div class="col-sm-6"> {{ else }} <div class="col-sm-12"> {{ end }}
+    <!-- takes up full section width if no badges/soft skills are specified -->
+    {{ if or (.softSkills) (.badges) }} <div class="col-sm-6"> {{ else }} <div class="col-sm-12"> {{ end }}
       <h3 class="p-1">{{ $author.name }}</h3>
       {{ if .designation }}
       <h5 class="p-1">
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank you @will-hou. Looks good to me.

PR #374, opened 2021-07-14
Rename zh.toml to tw.toml

HeiTang:
Rename it to be consistent with the flag name.


hossainemruz:
Language code doesn't have to match with country code.  You can use `falgOverwrite` to use different country flag.

Ref: https://github.com/hugo-toha/hugo-toha.github.io/blob/a60de3379102c31d450f20d6d42d9292035dcca1/config.yaml#L78

HeiTang:
OK, I didn't notice. Thanks @hossainemruz

Issue #375, opened 2021-07-15
SVGs are not supported in the icons for project card


jstehn:
## Expected Behavior
Set image to a card to be an svg file. It loads into the page 

## Current Behavior
Throws error when Hugo attempts to compile due to the `fit` command.

## Possible Solution
Add check for SVG file types. If it is not an SVG, then run the fit command

## Steps to Reproduce

1. Set logo value for a project to an SVG
2. Attempt to compile using Hugo
 
### Screenshots
![svg_error](https://user-images.githubusercontent.com/7625470/125739802-4c62adf3-c96c-4eac-b562-d3f54e42d250.png)

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

## Detailed Description
Support SVGs for logos in the Projects card

PR #376, opened 2021-07-15
Added svg support for project icons

jstehn:
### Issue
Fixes: https://github.com/hugo-toha/toha/issues/375

### Description

Handles SVGs for logos in the project card template by only calling the fit function if the icon is not an SVG

### Test Evidence

image


hossainemruz (approved):
Thank you @jstehn.

PR #383, opened 2021-07-29
Translate chinese

Chow-Ray:
Improve simplified Chinese translation, modify simplified Chinese and traditional Chinese file names。


hossainemruz (approved):
Thank you @Chow-Ray

PR #397, opened 2021-08-16
Fix multiple active menu in taxonomy

hossainemruz:
Signed-off-by: hossainemruz <hossainemruz@gmail.com>

cc: @Rihoj


Rihoj:
Thank you for fix that bug!

Issue #35, opened 2020-07-08
Newsletter functionality


oscarmrom:
Hello, 

I wanted to understand the functionality of the newsletter subscription in the footer:

![image](https://user-images.githubusercontent.com/26697231/86928745-f7389f80-c102-11ea-9029-be41b1895f0d.png)

How do you configure this, how is this expected to be used? I'm sorry if I missed the documentation on this. If you could link me that would be great!

hossainemruz:
Current, it doesn't work. I will add this functionality in future.

oscarmrom:
Okay, cool. I've commented it out from the footer on my end for the time being. Thank you!

hossainemruz:
@oscarmrom You can now hide the newsletter section using `config.yaml`. Ref: #48

PabloBiotech2020:
If I understand it corresctly, HUGO already auto generates RSS feeds for evert website, so couldnt this be done in a super cool, easy and standards compliant way if you simply redirected users to a site such as blogtrotter that notifies users on every new RSS item? I know, it wouldnt be the same as a full fledged newsletter but it would suffice to keep people informed of your blog changes, and, frankly, I think if people really need a real, ful fledged newsletter they would already be using something of the likes of mailchimp

MarionMoseby:
If I understand it corresctly, HUGO already auto generates RSS feeds for every website, so couldnt this be done in a super cool, easy and standards compliant way if you simply redirected users to a site such as blogtrotter that notifies users on every new RSS item? I know, it wouldnt be the same as a full fledged newsletter but it would suffice to keep people informed of your blog changes, and, frankly, I think if people really need a real, ful fledged newsletter they would already be using something of the likes of mailchimp

hossainemruz:
Added support for [mailchimp](https://mailchimp.com/) in #308.

PR #402, opened 2021-08-28
Add Blogtrottr as newsletter provider

MarionMoseby:
### Issue
 Might be related to issue  #35 (Newsletter functionality), although already solved

### Description

Hi! I was thinking of adding blogtrottr support to the theme's newsletter functionality, so that the default box works and is not just a dummy form. Blogtrottr is an RSS -> email service that automatically scans an RSS feed in search of updates from time to time and sends these updates in the form of an email to the user. As it works out of the box, it wouldnt need to get configured to get going, as is necessary, for example, with mailchimp.

Of course, it could be added as just another option by adding an else if statement; I just thought it was neater this way.

### Test Evidence

blogtrottr
 
At work on my personal website. As it is not live, no feed is detected (localhost only as I used absLangURL). Note that you can subscribe to a different feed for each language in the website

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -121,13 +121,13 @@
                 <button type="submit" class="btn btn-info">{{ i18n "submit" }}</button>
               </form>
             {{ else }}
-             <!-- a dummy form -->
-              <form>
-                <div class="form-group">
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank  you @MarionMoseby.

Rihoj:
@MarionMoseby I love this change! Awesome work!

PR #403, opened 2021-08-29
Added new 'GitForge' param for non GitHub repos

MarionMoseby:
### Issue
I use a gitea repository; with the current setup, there is no way to use it with toha

### Description

Makes the"Improve this page" button compatible with gitea and gitlab repos, while keeping backwards compatibility, I also tried to establish "main" as the default branch by doing: 
`
{{ if site.Params.GitBranch }}
  {{ $GitBranch = site.Params.GitBranch }}
{{ else }}
    {{ $GitBranch = main }}
{{ end }}

`

But I couldnt get it to work

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -71,7 +71,13 @@
         <!--- Improve this page button --->
         {{ if site.Params.GitRepo }}
           <div class="btn-improve-page">
-              <a href="{{ site.Params.GitRepo }}/edit/{{ site.Params.GitBranch }}/content/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
+            {{ if ( eq site.Params.GitForge "gitlab" ) }}
+              <a href="{{ site.Params.GitRepo }}-/edit/{{ site.Params.GitBranch }}/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
+            {{ else if ( eq site.Params.GitForge "gitea" ) }}
+              <a href="{{ site.Params.GitRepo }}_edit/{{ site.Params.GitBranch }}/content/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
+            {{ else }} <!--- Make Github-style the default -->
+               <a href="{{ site.Params.GitRepo }}/edit/{{ site.Params.GitBranch }}/content/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
+            {{ end }}
                 <i class="fas fa-code-branch"></i>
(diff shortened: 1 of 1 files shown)


layouts/_default/single.html
```diff
@@ -71,7 +71,13 @@ <h1>{{ .Page.Title }}</h1>
         <!--- Improve this page button --->
         {{ if site.Params.GitRepo }}
           <div class="btn-improve-page">
-              <a href="{{ site.Params.GitRepo }}/edit/{{ site.Params.GitBranch }}/content/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
+            {{ if ( eq site.Params.GitForge "gitlab" ) }}
+              <a href="{{ site.Params.GitRepo }}-/edit/{{ site.Params.GitBranch }}/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
```
hossainemruz, line 75: Should it be `"{{ site.Params.GitRepo }}/-/edit/` instead of `"{{ site.Params.GitRepo }}-/edit/` ?

layouts/_default/single.html
```diff
@@ -71,7 +71,13 @@ <h1>{{ .Page.Title }}</h1>
...
+            {{ else if ( eq site.Params.GitForge "gitea" ) }}
+              <a href="{{ site.Params.GitRepo }}_edit/{{ site.Params.GitBranch }}/content/{{ .File.Path }}" title="{{ i18n "improve_this_page" }}" target="_blank" rel="noopener">
```
hossainemruz, line 77: Should it be `{{ site.Params.GitRepo }}/_edit/{{ site.Params.GitBranch }}` instead of `{{ site.Params.GitRepo }}_edit/{{ site.Params.GitBranch }}` ?

hossainemruz (changes requested):
Thank you @MarionMoseby . Can you please acknowledge the comments?

layouts/_default/single.html
MarionMoseby: Hum I see, I believe the problem is my GitRepo param is styled as https://codeberg.org/FlyingFlamingo/pablomarcos.me/ instead of  https://codeberg.org/FlyingFlamingo/pablomarcos.me, so the {{ site.Params.GitRepo }} makes the / redundant. I guess for backwards compatibility I should add this change, even though I dont know which is more common, to have the final / or not

layouts/_default/single.html
MarionMoseby: Same here really

MarionMoseby:
Also, do you see if there is a way of making main the default Gitbranch? So that it does not have to be specified for most projects

MarionMoseby:
Everything should be ready now :p

hossainemruz (approved):
Looks good now. Thank you @MarionMoseby

PR #409, opened 2021-08-29
Update es.toml

MarionMoseby:
Updated after asking for permission in a question :p


hossainemruz (approved):
Thank you @MarionMoseby

PR #427, opened 2021-09-29
Update Simplified Chinese and Traditional Chinese.

Chow-Ray:
Signed-off-by: Ray <Chow-Ray@outlook.com>

#### Update Simplified Chinese and Traditional Chinese.


hossainemruz (approved):
Thank you @Chow-Ray

PR #436, opened 2021-10-17
Fixed section landing position on navigation #154

### LINKED ISSUE #154: Selecting sections from header menu but lands at different positions.
## Expected Behavior
The position should be fixed for the start of every section, whenever we select any section from header menu.

## Current Behavior
When we select Experience section, it goes to a position and then shifts up, some times it doesn't (which is correct.) 

## Steps to Reproduce

1. Select Experience section and notice the scroll, it goes to the correct position and shifts up.
2. Try randomly scrolling and going back to Experience section, sometimes it goes to the right position , sometimes it shifts up.

Have a look in the  gap between menu & Experience title.

### Screenshots

Correct position. (Correct gap between menu and Title)

image

Wrong position. (After the section shifts up, the gap between menu and Title reduces)

image

### Browsers Affected

- [ x] Chrome

- [x ] Edge
 &  All chromium based browser.

reyhhan:
### Issue
Fixes #154 

Re: When we select some sections from the menu, it slightly shifts up and does not position correctly.

### Description

In each section's `.html` file I added the `sectionID` to a new span tag that resides inside the section header tag. In the `.css` file, to the span tag I added a `margin-top` and `padding-bottom` approximately the size of the fixed header.

### Test Evidence

screen recording

Tested on Chrome, Firefox, IE

### DIFF
--- a/.github/workflows/autoprefixer.yml
+++ b/.github/workflows/autoprefixer.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout to latest commit
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v2.3.5
 
       - name: Setup Node
         uses: actions/setup-node@v2.4.1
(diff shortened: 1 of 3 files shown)


hossainemruz:
Thank you @reyhhan . I will take a look soon.

PR #471, opened 2021-12-02
Footer contact info: Icons and linking

tifrel:
IMO this makes the contact info in the footer:

- Tidier, because the FA icons have a fixed width
- More user-friendly, as clicking a link is always easier than copy-pasting,
  searching, etc.

It's also easily extendable by slapping on another `else if`.


hossainemruz:
Thank you @tifrel. I will review soon.

hossainemruz (approved):
@tifrel Updated the code a little bit. Now, the links open in a new tab. Also, removed an unnecessary `{{end}}`.

PR #481, opened 2021-12-26
Add LaTeX support

MarionMoseby:
Added katex scripts and css to scripts and header partials, respectively. Sorry I believe there are unnecesary extra commits maybe? Not sure what has happened


hossainemruz:
Thank you @MarionMoseby for the PR. The PR looks good. However, I think we should try to avoid using CDN whenever possible. Can you please include those CSS and JS files in static directory and use them from there?

hossainemruz:
@MarionMoseby I just noticed that LaTeX support is already included in the theme. You just have to add `math: true` in the post frontmatter.

Sample post: https://toha-guides.netlify.app/posts/writing-posts/math/
Markdown for the post: https://raw.githubusercontent.com/hugo-toha/guides/main/content/posts/writing-posts/math/index.md

MarionMoseby:
WoW, we have such a great team! Sorry for not looking through this thoroughly before adding a PR haha. 

Ok so, I have been tinkering with the code, and the problem seems to be that katex does not support inline latex using standard $...$ delimiters by default. This is because "dollar signs appear too often in non-mathematical settings, which could cause some text to be treated as mathematics unexpectedly" ([according to MatJax](https://docs.mathjax.org/en/latest/input/tex/delimiters.html), latex is bad at explaining this. So: would it be ok for me to modify this PR to:

- Update the Katex scripts from v0.11.1 to v0.15.1
- Make them local, instead of served over CDN
- Add a "renderwith$" parameter to math: yes so that $..$ rendering can be used

hossainemruz:
@MarionMoseby Sure, go on. Please make the changes in [math.html](https://github.com/hugo-toha/toha/blob/main/layouts/partials/math.html) file instead of the header and scripts. This will ensure that these scripts are not included on the pages where it is not necessary.

MarionMoseby:
We also have the following CDNs:
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Muli:wght@300;400;500;600">
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.11.2/css/all.min.css" />
Do you wish me to make them local too?

hossainemruz:
Yes. Please make all LaTex related CDN local.

MarionMoseby:
Sorry, I wrote the CDNS but markdown hid them haha. I meant the AJAX and google fonts CDNs, which are also rendered as CDN in the scripts and header sections; do you want me to make those local too?

hossainemruz:
Let's keep this PR LaTex related only. You can create another PR making other CDN resources local.

hossainemruz (approved):

PR #486, opened 2022-01-03
Fix Nav-Bar when link to section of page

### LINKED ISSUE #485: Nav-Bar is not displayed correctly when using a link that contains a section
## Expected Behavior
When visiting the page via a link that references a section (such as [https://hugo-toha.github.io/#skills](https://hugo-toha.github.io/#skills)) the nav-bar should be displayed correctly as shown in below screenshot.

Screenshot from 2022-01-02 19-55-44

## Current Behavior
The nav-bar is not displayed correclty. The current behavior is shown in below screenshot.

Screenshot from 2022-01-02 19-55-24

## Possible Solution

This code should be extracted in its own function and be called on page-load as well as on scroll (currently it is only called on scroll event)

https://github.com/hugo-toha/toha/blob/main/static/js/navbar.js#L11-L37

Please let me know if this would be an acceptable solution, in that case I will create a pull-request.

## Steps to Reproduce

Open this link  [https://hugo-toha.github.io/#skills](https://hugo-toha.github.io/#skills)

### Browsers Affected

Most likely all of them, tested in Firefox and Chromium.

t-muehlberger:
### Issue
Fixes #485 

### Description

Update nav-bar on page-ready as well as on scroll (rather than only on scroll event)

### Test Evidence

Everything seems to work on this page. 
image

### DIFF
--- a/static/js/navbar.js
+++ b/static/js/navbar.js
@@ -1,5 +1,36 @@
 "use strict";
 
+const updateNavBar = () => {
+  if ($(document).scrollTop() > 40) {
+    $('#top-navbar').removeClass('initial-navbar');
+    $('#top-navbar').addClass('final-navbar shadow');
+
+    $('#navbar-toggler').removeClass('navbar-dark');
+    $('#navbar-toggler').addClass('navbar-light');
+
+    // get the main logo from hidden img tag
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank you @t-muehlberger. Looks good to me.

PR #492, opened 2022-01-12
Localize post date

stueja:
### Description

The post date will be displayed in its localized version even if the page is translated.

### Test Evidence

Bildschirmfoto 2022-01-12 um 11 45 32


hossainemruz:
Thank you @stueja

PR #502, opened 2022-01-17
Update experience-info.html

stueja:
added .html to multiple-positions

### Issue
experience-info.html references file using incomplete filename

### Description
added .html to multiple-positions


hossainemruz (approved):
Thanks @stueja

PR #505, opened 2022-01-19
fix: navigation with hidden title

### LINKED ISSUE #494: Navigation not working when section title is hidden
## Expected Behavior
Even when the section title is hidden, a click onto the navbar should scroll to the corresponding section.

## Current Behavior
If the section title is hidden (`hideTitle: true` in e. g. `data/en/sections/experiences.yaml`), a click onto "Experiences" in the navbar does nothing.

## Possible Solution
Make the anchor independent from the title.

## Steps to Reproduce

1. set `hideTitle: true` in `data/en/sections/experiences.yaml`
2. run `hugo server -D`
3. open the webpage
4. click on "Experiences" in the navbar

### Browsers Affected

- [ ] Chrome
- [ x ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11
- [ x ] Safari 15

stueja:
### Issue
Navigation didn't work when section title was hidden. Fixes #494

### Description
Implemented if conditional. If `hidetitle` is true, `display: none` will be set. Thus, the `.section.id` to navigate to is still in the html to be navigated to.

### DIFF
--- a/layouts/partials/sections/accomplishments.html
+++ b/layouts/partials/sections/accomplishments.html
@@ -6,9 +6,12 @@
 <div class="container-fluid anchor pb-5 accomplishments-section">
   {{ if not (.section.hideTitle) }}
     <h1 class="text-center">
-      <span id="{{ $sectionID }}"></span>{{ .section.name }}</h1>
+        <span id="{{ $sectionID }}"></span>{{ .section.name }}</h1>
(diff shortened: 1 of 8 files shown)


hossainemruz (approved):
Thank you @stueja

PR #513, opened 2022-01-27
feat: customMenus in footer

stueja:
Show customMenu links also in footer.
In `config.yaml`,
 - `params.footer.enable` is `true` to show the footer at all
 - `params.footer.navigation.enable` is `true` to show navigation items
   in the footer
 - `params.footer.navigation.customMenus.enable` is `true` to show
   customMenu items

Example:
```
params:
  footer:
    enable: true
    navigation:
      enable: true
      customMenus:
        enable: true
```

In `data/<language>/site.yaml`,
 - `customMenus` must contain at least one entry with `showOnFooter:
   true`

Example:
 ```
 customMenus:
   - name: Imprint
     url: posts/imprint
     showOnFooter: true
 ```

PR #514, opened 2022-01-27
feat: configurable custom navbar items

stueja:
The customMenu item from `data/<language>/site.yaml` will only appear if
its property `showOnNavbar` is `true`.

Example:
```
customMenus:
  - name: Imprint
    url: posts/imprint
    showOnNavbar: false
```

### DIFF
--- a/layouts/partials/navigators/navbar.html
+++ b/layouts/partials/navigators/navbar.html
@@ -115,9 +115,11 @@
           </li>
         {{ end }}
         {{ range $customMenus }}
-          <li class="nav-item">
-            <a class="nav-link" href="{{ .url }}">{{ .name }}</a>
-          </li>
+            {{ if .showOnNavbar }}
+              <li class="nav-item">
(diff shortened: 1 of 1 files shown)


hossainemruz:
@stueja Can you please explain why it is necessary?

stueja:
It might make more sense to see this change in conjunction with #513. 
If you want to see a custom url "Imprint" in the footer but not in the navbar, you'd now have the chance to.
Without this change, all or no custom urls would appear.

hossainemruz:
Ok. I got it.

Btw, we should not change the default behavior. By default, a nav item should be visible in both top navbar and footer. Then, user can use special configuration to hide it from the navbar. Otherwise, it will breaks all the existing user's site.

We can use something like this:
```yaml
customMenus:
  - name: Imprint
    url: posts/imprint
    hideFromNavbar: true
```

stueja:
That's a good point! I would change it accordingly in the PR for the footer then.

hossainemruz (approved):

Issue #515, opened 2022-01-27
Lines in job timeline don't match up with, worsens with increased zoom.


kevinlinxc:
## Expected Behavior
Lines should match up.

## Current Behavior
At 120%, there's two visible cracks:
![image](https://user-images.githubusercontent.com/44324811/151436191-84edac59-d761-495a-a3a6-7a77e3724238.png)
At 100% zoom, there's only 1 crack:
![image](https://user-images.githubusercontent.com/44324811/151436314-a2602160-dceb-4e52-99e0-97d8e478ada7.png)
These only happen for some of the turns, not all.

## Steps to Reproduce
Visit https://kevinlinxc.github.io/ and scroll down to a job section.

### Browsers Affected
Only viewed in Firefox and Chrome, appears in both

## Detailed Description
The lines don't match up in the timeline. The corner piece appears to be a separate HTML element, and maybe there is rounding somewhere in the math that makes it not line up.

hossainemruz:
@kevinlinxc It should be fixed by #516. If it does not, feel free to re-open.

kevinlinxc:
![image](https://user-images.githubusercontent.com/44324811/151842821-5f48000c-1bed-40c4-b11c-84e52ee90a05.png)
This is what I see now, only on the first turn of the timeline. Viewable on https://kevinlinxc.github.io/
Checked Chrome and Firefox, happens on both.

hossainemruz:
Yes. That PR made it worse. I did not tested the PR before merging. I have now reverted back the change. In this commit: https://github.com/hugo-toha/toha/commit/69132a1dae5912f7898300cbbe27bc329ae334d0

hossainemruz:
I don't think we can fix it for every zoom level. The corner radius don't align with the horizontal/vertical lines for different zoom levels.

kevinlinxc:
For me, there's something wrong at every zoom level

hossainemruz:
Yes. I saw that. Not sure why it is happening.

PR #516, opened 2022-01-30
add calculation to line up lines

Rihoj:
### Issue
Hopefully fixes hugo-toha/toha#515

### Description

Since the element `.vertical-line-left-adjustment::after ` had a calculated value of 50% - 3px I added that to the corner object as well.

### Test Evidence

image


hossainemruz (approved):
Thank you @Rihoj

hossainemruz:
It did not solve the issue.

Issue #539, opened 2022-02-27
Experience `Responsibilities:` visible when there are no responsibilities


ivankatliarchuk:
## Expected Behavior

The file `layouts/partials/sections/experiences/single-position.html` render `Responsibilities:` even when `experience` config does not contain `responsibilies` for an experience

Example `experiences` file affected

```yml
# section information
section:
  name: Experiences
  id: experiences
  enable: true
  weight: 3
  showOnNavbar: true
  # Can optionally hide the title in sections
  # hideTitle: true 

# Your experiences
experiences:
- company:
    name: Example Co.
    url: "https://www.example.com"
    location: Dhaka Branch
    # company overview
    overview: Example Co. is a widely recognized company for cloud-native development. It builds tools for Kubernetes.
  positions:
  - designation: Senior Software Engineer
    start: Nov 2019
```

## Current Behavior
`Responsibilities:` to  present

## Possible Solution
File `toha/layouts/partials/sections/experiences/single-position.html` 

Current
```yml
    <!-- Add company overview -->
    <p>{{ .company.overview | markdownify }}</p>
    <!-- Add the responsibilities handled at this position -->
    <h6 class="text-muted">{{ i18n "responsibilities" }}</h6>
    <ul class="justify-content-around">
    {{ range $position.responsibilities }}
        <li>{{ . | markdownify }}</li>
    {{ end }}
    </ul>
```

Fix
```yml
    <!-- Add company overview -->
    <p>{{ .company.overview | markdownify }}</p>
    <!-- Add the responsibilities handled at this position -->
    {{ if $position.responsibilities }}
    <h6 class="text-muted">{{ i18n "responsibilities" }}</h6>
    <ul class="justify-content-around">
    {{ range $position.responsibilities }}
        <li>{{ . | markdownify }}</li>
    {{ end }}
    </ul>
    {{ end }}
```

## Steps to Reproduce

1. Comment lines https://github.com/hugo-toha/hugo-toha.github.io/blob/3201586462e5623b3f6efbe0e8b9fe88f77fe7ab/data/en/sections/experiences.yaml#L33-L36
2. Start the example

### Screenshots
Bug
![before-resp-empty](https://user-images.githubusercontent.com/5395690/155899999-6eacda1d-ed3c-438c-969e-c318733a8a0d.png)

![before-resp-commented](https://user-images.githubusercontent.com/5395690/155900000-43dafa5b-d718-47a8-a2fb-84daf341eda8.png)

With the fix
[...]

PR #540, opened 2022-02-27
Don't show `Responsibilities` header, if there aren't any

ivankatliarchuk:
### Issue

Fixes https://github.com/hugo-toha/toha/issues/539

### Description

Block `{{ if $position.responsibilities }}` is responsible for rendering `Responsibilities:` only when there is at least single responsbility

### Test Evidence
Run locally

after-with-the-fix


hossainemruz (approved):
Thank you @ivankatliarchuk .

PR #542, opened 2022-02-28
add rss feed

Rihoj:
### Issue
Fixes #533 

### Description

Add a modified rss feed. Based off https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_default/rss.xml

Uses RegularPagesRecursive
(Requires Hugo v0.68.0)

Added an if statement to not include the search pages (based off the URL field)

### Test Evidence
<details><summary>posts/index.html</summary>

```
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<rss xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
<channel>
<title>Posts on James Ray's Blog</title>
<link>http://localhost:1313/posts/</link>
<description>Recent content in Posts on James Ray's Blog</description>
<generator>Hugo -- gohugo.io</generator>
<language>en-us</language>
<atom:link href="http://localhost:1313/posts/index.xml" rel="self" type="application/rss+xml"/>
<item>
<title>Gurkha Rare Breed</title>
<link>http://localhost:1313/posts/pipes/tobacco/gurkha-rare-breed/</link>
<pubDate>Sun, 27 Feb 2022 22:31:00 -0500</pubDate>
<guid>http://localhost:1313/posts/pipes/tobacco/gurkha-rare-breed/</guid>
<description>Tobacco: Red Virginia, Turkish, Burley, Latakia, and Perique Style: Aromatic Room Note: 4 - Distinct Cut: Ribbon Cut Top Flavor: Misc Flavor Intensity: 2 - Mellow Review 5/5 stars Man does this smell take me back! I grew up around a circus, and one whiff of this really shoots me back to the smell of that hay. The flavor, and strong nicotine content really sets this one apart for me.</description>
</item>
<item>
<title>Designing the Front End Part 1</title>
<link>http://localhost:1313/posts/video-streamer/frontend/design/</link>
<pubDate>Wed, 09 Feb 2022 22:15:00 +0000</pubDate>
<guid>http://localhost:1313/posts/video-streamer/frontend/design/</guid>
<description>Choosing my framework So I am a gluten for punishment some times. I do not enjoy front end designing at all. I also don&rsquo;t take the time to layout a workflow, or any form of design structure. So where do you start if you don&rsquo;t even know how your site is going to behave? How do you know if you are going to get the right framework? IDK. I just grabbed the one I like the most.</description>
</item>
<item>
<title>Getting Started</title>
[...]

PR #548, opened 2022-03-01
Auto build

Rihoj:
### Issue
#542 caused an issue with builds. This could have been spotted if builds were turned on.

### Description
This workflow runs the basic build command just to confirm that the PR will actually build.


hossainemruz (approved):
Thank you @Rihoj

PR #551, opened 2022-03-06
Add optional mermaid shortcode

### LINKED ISSUE #550: Mermaid support
How to integrate mermaid plugin

Rihoj:
### Issue
Fixes #550

### Description

Adds support for [Mermaid-js](https://mermaid-js.github.io/mermaid/#/)

### Warning
Styling has not been worked out, and there may be some odd looks. Check the linked issue #550 for an example.


hossainemruz:
@Rihoj Updated the shortcode. Now, it uses page level configuration instead of global mermaid configuration. In order to enable mermaid for a page, you have to set `mermaid: true` in the page frontmatter. Now, mermaid script will load only in those pages.

Regarding the style issue of Git graph. It seems to be an issue of mermaid-js itself. We have no control of the generated svg. We can use custom theme but it will be lot of hassle. So, I have added an optional `background` parameter to the shortcode. User can set this to `black` when there is any readability issue.

Let me know, what you think.

hossainemruz (approved):

PR #565, opened 2022-03-23
Add finnish translation

FunctionalHacker:
Here is the finnish translation. I was not able to translate "at" and "out of" since finnish is very different from english and I would need to have some context on how it is used. I've left them as english for now.


hossainemruz:
Thank you @FunctionalHacker for the PR. Can you please add Finish in the [Available Translations](https://github.com/hugo-toha/toha#available-translations) section of README.md file?

hossainemruz (approved):

PR #571, opened 2022-03-26
Enable pages in counter.dev

salastro:
### Issue

Counter.dev does not report the visited pages inside the site
### Description

The added line is a workaround to solve this issue. However, it is not perfect because it will stop working after some time, but since there is no other solution I was able to find, this is the best that can be done in the mean time.

More about it on https://github.com/ihucos/counter.dev/issues/37.

### Test Evidence

None, I made this in a rush.


MarionMoseby:
@salastro why will it stop working after some time?

layouts/partials/analytics.html
```diff
@@ -22,7 +22,8 @@
                     screen: screen.width + "x" + screen.height,
                     user: "{{ .id }}",
                     utcoffset: "1",
-                    })
+                    }),
+                    { referrerPolicy: "no-referrer-when-downgrade" }
```
hossainemruz, line 26: This should be a configurable field with default.

hossainemruz (approved):

salastro:
> @salastro why will it stop working after some time? How much time?

I am not exactly sure when the support for this feature will be dropped by browsers. The alert just says `Referrer Policy: Less restricted policies, including ‘no-referrer-when-downgrade’, ‘origin-when-cross-origin’ and ‘unsafe-url’, will be ignored soon for the cross-site request` which is not specific at all. Furthermore, I was not able to find anything helpful through my search around the internet.

layouts/partials/analytics.html
salastro: I am not used to modifying Hugo themes, but I will try to do it anyway. 

> with default.

What should the default be? With or without the `referrerPolicy`?

hossainemruz: Ok. I can work on making it configurable. I have check to decide the default value.

salastro:
I have modified the template to check the variable `noReferrerWhenDowngrade` under `counterDev` to decide whether to add the `{ referrerPolicy: "no-referrer-when-downgrade" }` section. The default is not to add it since this is the default for counter.dev; to enable it you have to make the value of the variable `true`.

layouts/partials/analytics.html
```diff
@@ -22,7 +22,10 @@
                     screen: screen.width + "x" + screen.height,
                     user: "{{ .id }}",
                     utcoffset: "1",
-                    })
+                    }),
+		    {{ if .noReferrerWhenDowngrade }}
```
hossainemruz, line 26: It will be better to use a variable `referrerPolicy` instead of `noReffererWhenDowngrade`. Then, user will have option to use other policies too.

We can just do:
```
{{ if .referrerPolicy }}
  { referrerPolicy: "{{ .referrerPolicy}}" }
{{ end }}
```

hossainemruz (changes requested):

layouts/partials/analytics.html
salastro: I believe this is the correct form for such a task. Is there anything else that I can do (or something that I have done wrong)?

hossainemruz: Thank you. Now, it looks good to me.

hossainemruz:
It will be really helpful if you can add this field reference here: https://github.com/hugo-toha/guides/blob/main/content/posts/analytics/index.md#counterdev

PR #575, opened 2022-04-01
Post specific togglable TOC

### LINKED ISSUE #574: Disabe TOC if there is no headings in the post

If there is no suitable headings in the post, then do not show the table of contents.

### Describe alternatives you've considered

Making it possible to disable in post configuration (`TOC: false` only for this post).

### Additional context

Sometimes, I write super simple stuff that does not need headings, so the TOC looks very annoying since it's completely empty

salastro:
### Issue

Fixes #574 

### Description

If both site-configured TOC and post-configured TOC is set to true, then enable TOC, disable otherwise. The default of post-configured TOC is `true`.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -179,7 +179,7 @@
 
 {{ define "toc" }}
   <section class="toc-section" id="toc-section">
-    {{ if site.Params.enableTOC }}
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank you @salastro

Issue #59, opened 2020-08-04
Add dark mode


hossainemruz:
https://dev.to/akhilarjun/one-line-dark-mode-using-css-24li

alex-bezek:
@hossainemruz FWIW this is what it does to my site
Before
![image](https://user-images.githubusercontent.com/8029578/96395551-e7fdcf00-1192-11eb-964b-ea8f524cd079.png)
![image](https://user-images.githubusercontent.com/8029578/96395007-75d8ba80-1191-11eb-9d2a-42c363e81030.png)

After
![image](https://user-images.githubusercontent.com/8029578/96395724-622e5380-1193-11eb-8967-6911d61757b2.png)
![image](https://user-images.githubusercontent.com/8029578/96394997-6eb1ac80-1191-11eb-9876-7eedf5bf5a78.png)

I think largely it does the job well but there are a few sets of text that need some attention in order for their contrast ratio to be high enough. i assume if we figure those out we could easily turn this on with a feature flag in the config.yml to make this passive?

yeganathan:
Hey @hossainemruz can you please tell me for darkmode in which CSS file should be the following code added,

```
html[theme='dark-mode'] {
    filter: invert(1) hue-rotate(180deg);
}
```

hossainemruz:
I am not sure yet @yeganathan. Probably, @alex-bezek can help.

yeganathan:
Thanks for the reply! @hossainemruz 
Let me check with @alex-bezek

alex-bezek:
@yeganathan18 if you add a css file to this location, it will get picked up and included in the build
https://github.com/alex-bezek/blog/blob/master/static/assets/css/style.css

yeganathan18:
@alex-bezek Thanks for the help :)

donfiguerres:
I'm experimenting with using DarkReader to enable dark mode in toha. My branch is [here](https://github.com/donfiguerres/toha/tree/dev-dark-theme).
So far, the result is pretty decent. Not sure if that's biased because I'm using the DarkReader browser plugin. I think a native CSS for dark mode is faster but still, this gets the job done with minimal effort.

![image](https://user-images.githubusercontent.com/12525903/119544484-53acb800-bdc4-11eb-9fd9-4b319ddc4be8.png)
![image](https://user-images.githubusercontent.com/12525903/119544712-8eaeeb80-bdc4-11eb-939a-2b45be2f3c70.png)

[...]

PR #595, opened 2022-06-03
Dark theme via DarkReader

donfiguerres:
### Issue

Closes https://github.com/hugo-toha/toha/issues/59

### Description

Use [DarkReader](https://darkreader.org/) as a darkmode provider.
I've added the option to use other theme color providers and hopefully toha's own dark theme in the future. This possible with the darkTheme.provider property. Below is an example on how to set-up the dark theme. I'm also planning to add this information to the toha documentation.

    # Dark theme toggle
    darkTheme:
      provider: darkreader
      enable: true
      default: dark

The drawback of this is that it is a lot slower than having your own dark theme because the theme is dynamically generated.

### Test Evidence

It's already live in my website https://donfiguerres.com/
image
image


layouts/partials/scripts.html
```diff
@@ -5,3 +5,10 @@
 <script type="text/javascript" src="{{ "/js/navbar.js" | relURL }}"></script>
 <script type="text/javascript" src="{{ "/js/plyr.js" | relURL }}"></script>
 <script type="text/javascript" src="{{ "/js/main.js" | relURL }}"></script>
+
+{{ if site.Params.darkTheme.enable }}
+{{ if eq site.Params.darkTheme.provider "darkreader" }}
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/darkreader@4.9.51/darkreader.js"></script>
```
hossainemruz, line 11: Let's use local file instead of CDN. Some people seems to be concerned with CDN. Include the file in js folder.

layouts/partials/navigators/theme-selector.html
```diff
@@ -0,0 +1,12 @@
+<li class="nav-item dropdown">
+<!-- This is a 'hack' for initializing the color scheme selection. See /js/darkmode.js -->
+<div id="themeInitialization" style="display: none;"
+    default-theme=" {{ site.Params.darkTheme.default }} "></div>
+<a class="nav-link dropdown-toggle"  href="#" id="themeSelector" role="button"
```
hossainemruz, line 5: Can we use icon instead of text? It will differentiate it from rest of the menus. Probably, something like this: https://tailwindcss.com/

hossainemruz (changes requested):
Hi @donfiguerres, Thank you for the amazing work. I have added few comments. Can you acknowledge them. Otherwise, looks good to me.

layouts/partials/navigators/navbar-2.html
```diff
@@ -47,6 +47,9 @@
       {{ if .IsTranslated }}
         {{ partial "navigators/lang-selector-2.html" . }}
       {{ end }}
+      {{ if site.Params.darkTheme.enable }}
```
hossainemruz, line 50: Should we rename `darkTheme` to `darkMode` ?

hossainemruz (changes requested):

donfiguerres:
> Hi @donfiguerres, Thank you for the amazing work. I have added few comments. Can you please acknowledge them. Otherwise, looks good to me.

Thanks for the feedback! I agree with your inputs. I'll work on these and push the changes to my branch.

donfiguerres:
Done with the changes and already live on my website at https://donfiguerres.com/ if you need to see how it looks.

I also did a quick test with the multilingual mode to make sure it doesn't conflict with the language dropdown menu.

image
image

layouts/partials/scripts.html
donfiguerres: Added a modified copy of darkreader.js.

layouts/partials/navigators/navbar-2.html
donfiguerres: Renamed this to darkMode.

hossainemruz:
Hi @donfiguerres, thanks again for the hard work. Looks good to me. I will test and merge soon.

hossainemruz (approved):

PR #603, opened 2022-06-12
Removed flag icon span if countryCode is null

GabrielJean:
### Issue
N/A

### Description

Added some logic to remove the flag icon span if flags are removed with flagOverwrites

### Test Evidence
Before:
image

After: 
image

Thanks 😃

### DIFF
--- a/layouts/partials/navigators/lang-selector-2.html
+++ b/layouts/partials/navigators/lang-selector-2.html
@@ -6,7 +6,9 @@
 <li class="nav-item dropdown">
     <a class="nav-link dropdown-toggle" href="#" id="languageSelector" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
       {{ $countryCode := partial "helpers/country-code.html" . }}
+      {{ if ne $countryCode nil  }}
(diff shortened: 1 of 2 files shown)


hossainemruz:
The two screenshot looks identical. Whats the difference? Also, why this is necessary? Why the country code will be nil?

GabrielJean:
The difference is the space between the two buttons that is left behind if the flag is empty. As of why, I personally prefer to not have the flags and I would like to allow this to be possible without the empty space. I know it's s very small detail but since it doesn't break anything I figured I should propose it upstream.

image

Thanks

pboling (approved):
This looks great!  Love not having the blank space, and not having to have any flag!

hossainemruz:
@GabrielJean Then, we should have a better way to specify that you don't want to show the flag instead of this workaround. Probably a field `showFlags: false` which will be default to `true`.

GabrielJean:
@hossainemruz I made the changes as you suggested, does that work for you?

hossainemruz (approved):

GabrielJean:
Sounds good! 

Thanks

PR #639, opened 2022-09-02
Local google font

### LINKED ISSUE #630: Is there a way to not use Google fonts and add privacy policy?
### Question
I would like to use a different font system other than Google fonts, partly due to privacy reasons and to be compliant with privacy laws. According to Cookiebot.com, my implementation of the theme is not compliant with ePrivacy Directive (ePR).  Is it also possible to add a dedicated privacy policy and copyright disclaimer? Thanks for the amazing theme! 👍

LNA-DEV:
### Issue
[Is there a way to not use Google fonts and add privacy policy?](https://github.com/hugo-toha/toha/issues/630)

Fixes #630 

### Description

Added the google font locally as well as the fonts.css file. This is very improtant for privacy reasons. In some countrys such as Germany the theme would not be legal to use. 
(Google Font Muli was renamed to Mulish)

### Test Evidence

Check the network inspector


hossainemruz:
Thank you for @LNA-DEV. I will review and merge soon.

hossainemruz (approved):

Issue #399, opened 2021-08-19
Flag 404 bug


Rihoj:
## Expected Behavior
Either don't try to load the files or include them in the theme.

## Current Behavior
Tries to load svgs for the flags, that don't exist

## Possible Solution
Check for files first, or just include them in the theme

## Steps to Reproduce
1. If you open the Devtools network or console you will be able to see 258 errors of 404s https://rayprogramming.com

### Screenshots
![image](https://user-images.githubusercontent.com/1520296/130113680-11cbad26-d77e-4804-a897-fe5c9671122b.png)

### Browsers Affected
Untested on anything other than chrome
- [ x ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

Nirzak:
Hi, I don't see any such errors on my end. May be you should upgrade your themes to the latest one.
![Screenshot (3770)](https://user-images.githubusercontent.com/11460645/136657053-29f981a5-90e3-41ab-9b6c-d4ae328582ec.png)

hossainemruz:
This should be fixed now. Please, try upgrading to the latest release. Feel free to reopen if not fixed.

PR #658, opened 2022-10-05
fix: fixed http 404 for flags required by darkreader

TheDarthMole:
### Issue
[The HTTP 404 flags bug](https://github.com/hugo-toha/toha/issues/399) has re-appeared, this is due to DarkReader requiring flags to be accessible in `/flags/1x1/`

### Description

I moved flags from [this repository](https://github.com/lipis/flag-icons/tree/main/flags/1x1) to the `static/flags/1x1` directory so that they are served to the end user.

### Test Evidence
Before fix:
Before image

After fix:
After image


hossainemruz:
Thank you @TheDarthMole

hossainemruz (approved):

PR #662, opened 2022-10-09
Improved german localization

BenjaminAdam:
"Guthaben" is used for banking loans.
Changed to "Punktzahl" and "Note" to
better reflect the german education
system.

The 404 error message was just
nonsense. Changed it to correct
translation of english version.


hossainemruz (approved):
Thank you @BenjaminAdam

Issue #663, opened 2022-10-11
Arrow not centered


LNA-DEV:
## Expected Behavior

Arrow should be centered on the home page.

## Current Behavior

Arrow is slightly on the right.

## Possible Solution

Center the parent element.

## Steps to Reproduce

It happens on every site.

### Screenshots

![Screenshot from 2022-10-11 10-47-41](https://user-images.githubusercontent.com/77580596/195046493-b7b9d19e-e638-4e92-8bf0-3c5c855a593a.png)

PR #664, opened 2022-10-11
Centered Arrow

LNA-DEV:
### Issue

#663 

### Description

The arrow on the home page is not centered. It is slightly on the right.

I also centered the parent of the element which really centers the whole arrow.

### Test Evidence

For example you can take a look at the example page [https://hugo-toha.github.io/](https://hugo-toha.github.io/).

It works on my site [https://lna-dev.net](https://lna-dev.net).


hossainemruz (approved):
Thank you @LNA-DEV. Looks good to me.

PR #666, opened 2022-10-15
Only load flag CSS if showing flag icons

Cyb3r-Jak3:
### Issue
Currently, if you had `showFlags: false` the CSS file  `/css/flag-icon.min.css` is still loaded even though I am pretty sure it is not being used.

### Description

This PR disables the loading of the flag CSS file if `showFlags` is false.

### Test Evidence

Tested locally by setting `showFlags:` to false and not seeing the flag CSS then and then set to true and see it be loaded.


hossainemruz (approved):
Thank you @Cyb3r-Jak3

PR #670, opened 2022-10-15
Allow custom hero images Fixes #379

### LINKED ISSUE #379: Hero image and posts bug
Hello,

I followed this [article](https://toha-guides.netlify.app/posts/writing-posts/organizing/category/) but I faced some issues where I couldn't set a custom hero image except by doing the following.
1. The here image should be named here, so I can't use a custom name.
2. It should be in the post folder, and any other path will not work even `assets/images` and It will always fall back to the default hero image.
3. The post name should be `index.md` other than that it will show the default hero image even if the previous points are satisfied.

The other issue is that recent posts always show the following, if I have only one post.

Screen Shot 2021-07-24 at 5 11 44 PM

Please let me know If I'm doing something wrong.
Thanks in advance.

SrNetoChan:
### Issue

Fixes #379
### Description

Fixes issues that does not allow to set a custom hero for each post unless you have a bunch conditions met

It's based on a change proposed earlier but for some reason was deleted #661

### Test Evidence

With this change, I was able to put the hero image in the assets folder and set the hero parameter in the post pointing to 
`/images/name_of_the_file.png`

image


hossainemruz (approved):
Thank you @SrNetoChan

hossainemruz:
@eishundo This PR fixes the issue: https://github.com/hugo-toha/toha/pull/684

PR #676, opened 2022-10-19
Better Support if browser blocks GoatCounter

### LINKED ISSUE #675: AdBlockers can block GoatCounter js file

LNA-DEV:
### Issue

Fixes #675

### Description

By provisioning the js file localy the browser / ad blocker can be tricked

### Test Evidence

Just try it out with the change and you will see that file request is not blocked any longer. (Tested with the Brave Browser)


hossainemruz (approved):
Thanks @LNA-DEV

Issue #688, opened 2022-11-05
HTML Proofer complains that "alt" attributes are missing from img tags in theme-selector.html


eishundo:
## Expected Behavior

img tags should contain "alt" attributes to pass HTML Proofer tests.

## Current Behavior

HTML Proofer complains that "alt" attributes are missing from img tags generated by layouts/partials/navigators/theme-selector.html

```
  <img id="navbar-theme-icon-svg" src="{{ "/icons/moon-svgrepo-com.svg" }}" width=20>
    <img class="menu-icon-center" src="{{ "/icons/sun-svgrepo-com.svg" }}" width=20>
    <img class="menu-icon-center" src="{{ "/icons/moon-svgrepo-com.svg" }}" width=20>
    <img class="menu-icon-center" src="{{ "/icons/computer-svgrepo-com.svg" }}" width=20>
```

## Possible Solution

Submitting patch to add alt attributes to img tags.

```
  <img id="navbar-theme-icon-svg" src="{{ "/icons/moon-svgrepo-com.svg" }}" width=20 alt="Dark Theme">
    <img class="menu-icon-center" src="{{ "/icons/sun-svgrepo-com.svg" }}" width=20 alt="Light Theme">
    <img class="menu-icon-center" src="{{ "/icons/moon-svgrepo-com.svg" }}" width=20 alt="Dark Theme">
    <img class="menu-icon-center" src="{{ "/icons/computer-svgrepo-com.svg" }}" width=20 alt="System Theme">
```

Pull request submitted: https://github.com/hugo-toha/toha/pull/689

## Steps to Reproduce

Run "HTML Proofer" action as per: https://github.com/hugo-toha/hugo-toha.github.io/actions/workflows/html-proof.yaml

### Screenshots

Before patch:
![theme-selector-before](https://user-images.githubusercontent.com/116754122/200118574-c45753d5-8142-4537-9a0d-b5b9addd5b06.png)

## After patch:
![theme-selector-after](https://user-images.githubusercontent.com/116754122/200118572-b02e8f48-39f8-4a8d-b0a1-5d1c07d41e3e.png)

### Browsers Affected

N/A

## Detailed Description

Simply adding missing "alt" attributes to img tags.

PR #689, opened 2022-11-05
Add missing "alt" attributes to img tags in theme-selector.html

eishundo:
### Issue

#688 - HTML Proofer complains that "alt" attributes are missing from img tags in theme-selector.html

### Description

HTML Proofer complains that "alt" attributes are missing from img tags generated by layouts/partials/navigators/theme-selector.html

## Before patch:
theme-selector-before

## After patch:
theme-selector-after


hossainemruz (approved):
Thank you @eishundo

PR #703, opened 2022-11-12
Fix links of `posts`, `notes`, and `tags/%s` directories

hiroyaonoe:
### Issue
https://toha-guides.netlify.app/posts is going to be redirected to https://toha-guides.netlify.app/posts/ because `posts` is a directory. `notes` and `tags/%s` are also the same.  
There are such redirect links on some pages. For performance, it's not preferred to redirect them.

### Description
This PR correct `posts`, `notes` and `tags/%s` to `posts/`, `notes/` and `tags/%s/`

### Test Evidence
Their links are going to be redirected to the correct pages, so there is no change in their behavior.
For example, the `Posts` link at the navbar in https://toha-guides.netlify.app/ will change from https://toha-guides.netlify.app/posts into https://toha-guides.netlify.app/posts/, but Both destination pages will be the same.

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -24,7 +24,7 @@
         </form>
         <div class="sidebar-tree">
           <ul class="tree" id="tree">
-            <li id="list-heading"><a href="{{ "/posts" | relLangURL }}" data-filter="all">{{ i18n "posts" }}</a></li>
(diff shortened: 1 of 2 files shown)


hossainemruz (approved):
Thank you @hiroyaonoe

Issue #726, opened 2023-01-06
Brazilian Portuguese translation


igorr0driguez:
### Describe the solution you'd like
Adding Brazilian Portuguese translation. The European Portuguese one can work but "Home" button will be missing (because there's no translation "pt-br" for "Home"). Also, the theme will have some trouble running when there's only "pt-br" files. So I think adding a proper translation for Brazil is adequate.

### Additional context
I've already done it and I'm creating the pull request.

PR #727, opened 2023-01-06
Brazilian Portuguese translation

igorr0driguez:
### Issue

Fixes https://github.com/hugo-toha/toha/issues/726

### Description

I'm adding Brazilian Portuguese translation to Toha.
### Test Evidence

toha
homemissingtoha

Notice that the "Home" button is missing, because there's no proper pt-br translation files. Also, the theme is having some trouble accepting only "pt-br" files as a language. The reason there's a lot of missing things from the Menu there is because I removed them when I've taken it, the Menu is working just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ For more details about the features please visit [here](https://toha-guides.netl
 - Arabic (العربية)
 - Português
 - Català
+- Português Brasileiro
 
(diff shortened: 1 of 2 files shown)


hossainemruz (approved):
Thank you @igorr0driguez.

PR #747, opened 2023-04-06
fix cardHeader

Emily-0616:
Would it be better to center the preview image vertically?

The left side is after modification, the right side is before modification
image

The left side is after modification, the right side is before modification
image

The left side is after modification, the right side is before modification
image


hossainemruz (approved):
Thank you @Emily-0616.

Issue #754, opened 2023-04-24
Changing default color scheme in config.yaml has no effect


mmzeynalli:
## Expected Behavior

The default color scheme in config.yaml file is 'system', and it works okay. If you change it to dark, the default color scheme should be dark.

## Current Behavior

However, it does not depend to what you change, it stays in 'system' configuration when you load a page.

## Steps to Reproduce

1. Both in Windows and Debian, changing config.yaml, does not affect end result.
2. Node version: 18.16.0
3. Npm version: 9.6.5
4.

### Browsers Affected

- [ ] Chrome
- [x] Firefox
- [x] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

hossainemruz:
Fixed in #755

PR #755, opened 2023-04-24
Fixed changing default color scheme in config, affecting the result

mmzeynalli:
### Issue

[#754 Changing default color scheme in config.yaml has no effect](https://github.com/hugo-toha/toha/issues/754)

### Description
Upon debugging, I found that, in `/assets/scripts/features/darkmode/darkreader.js` file, `darkreader` object has attribute `defaultcolorscheme` not, `defaultColorScheme`, thus the following code always results in 'system' configuration.

`const defaultColorScheme = darkreader.defaultColorScheme || 'system'`

### Test Evidence
image


hossainemruz (approved):
Thank you @mmzeynalli

mmzeynalli:
My pleasure) Thanks for sharing such a wonderful template!

Issue #774, opened 2023-07-05
Math rendering with Katex not rendering


BernatBC:
## Expected Behavior
Katex should render math formulas.

## Current Behavior
The website does not render the formulas using Katex, shows plain text.

## Possible Solution
I don't know yet, I'll check this issue. If I can fix it, I'll do a Pull Request.

## Steps to Reproduce
Just go to https://hugo-toha.github.io/posts/markdown-sample/ and scroll down to `Math Rendering`. This issue also occurs to my website.

### Screenshots
![latex](https://github.com/hugo-toha/toha/assets/70479573/ee592542-a884-4389-9315-d913df5cadf8)

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

Haven't checked Edge, Safari, Explorer, probably happens the same.

richbai90:
Did you set math:true in your frontmatter?

```markdown
---
title: "my post"
date: 2023-07-25T16:11:23-06:00
draft: false
math: true
menu:
  sidebar: ...
---

BernatBC:
Of course I did. Is  not a matter of configuration as I had this kattex thing working before the major rewrite. The example website also has the same issue. Also, I have done a Pull Request Fixing this issue.

Thank you for trying to help :)

BernatBC:
As you can see in #775, I was wrong. This works perfectly.

PR #775, opened 2023-07-05
Fix: Math rendering with Katex not rendering (#774)

BernatBC:
### Issue
(https://github.com/hugo-toha/toha/issues/774)

### Description
Fixed katex rendering issue. There may be better solutions, but this one works.

### Test Evidence
image

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -18,3 +18,7 @@
 
   <link rel="icon" type="image/png" href="{{ $favicon }}" />
 {{end}}
+
+{{ if or .Page.Params.math .Site.Params.math }}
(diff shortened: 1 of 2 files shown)


hossainemruz:
Hi @BernatBC, math rendering actually works properly. All you have to do is to enable the feature in your `config.yaml` file. Set `enabled: true` here:
https://github.com/BernatBC/BernatBC.github.io/blob/c86682d1a57f9f89925a3f84b79b6940e53f8e60/config.yaml#L174

BernatBC:
Thank you so much for your help, I had overlooked it.
I'm sorry to have caused you inconvenience.

PR #782, opened 2023-07-28
Add possibility to show tags on post cards

BernatBC:
### Description
When enabling this feature in `config.yaml`
```yaml
tags:
      enable: true
      on_card: true
```
post cards in `Recent Posts`, `Posts`, `Tags`, etc. will contain the tags that are specified from that post.

### Test Evidence
Before:
before

After:
image

### Comments
- I was going to add it to `guides` but they are outdated, tell me if I have to add it to `guides` or to `hugo-toha.github.io` (as disabled, as tags are not enabled in this site).
- If you want, we can change the location of this setting, but I think under `tags:` in `config.yaml`  is the best location.
- Also, we can change the name of this setting.

### DIFF
--- a/assets/styles/layouts/list.scss
+++ b/assets/styles/layouts/list.scss
@@ -98,6 +98,22 @@ body.kind-section, body.kind-term {
     width: fit-content;
   }
   
+  .taxonomy-terms-card {
+    text-align: left;
+  }
+  .taxonomy-terms-card li {
+    font-size: 0.5em;
+    list-style-type: none;
+    display: inline-block;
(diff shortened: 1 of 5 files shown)


hossainemruz:
Hi @BernatBC! Thank you for the great PR. Can you please add the configuration sample under both `exampleSite/config.yaml` and also in `hugo-toha.github.io` repo?

BernatBC:
I have applied this changes. You can merge this PR whenever you want.

hossainemruz (approved):
Thank you @BernatBC

PR #790, opened 2023-08-25
fix dark mode: absolute to relative icon path

vincentDcmps:
### Issue
when use a base url with a folder dark mode icon get 404

### Test Evidence
Capture d’écran du 2023-08-25 16-27-08
Capture d’écran du 2023-08-25 16-26-37


hossainemruz (approved):
Thank you @vincentDcmps. I have done little modification to resolve conflict with upstream.

Issue #798, opened 2023-09-29
Navbar spy disappeared after updating to toha v4


daviddavo:
### Describe the solution you'd like

Having the navbar scroll spy again. Some of my sections are quite big and I'd like to use it.

### Describe alternatives you've considered

Not updating toha

daviddavo:
I fixed it, I'll make a PR

PR #799, opened 2023-09-29
Fixed navbar scroll spy (#798)

daviddavo:
### Issue
[](https://github.com/hugo-toha/toha/issues/798)

- https://github.com/hugo-toha/toha/issues/798

### Description

Followed guide in [bootstrap docs](https://getbootstrap.com/docs/4.6/components/scrollspy/#fat), and removed website url from anchors, but all browsers accept this and is common place to put just the hash-link.

### Test Evidence

Screenshot_2023-09-29_21-13-35


layouts/partials/navigators/navbar.html
```diff
@@ -102,7 +102,7 @@
                   {{ if and (.section.enable) (.section.showOnNavbar) }}
                     {{ $sectionCount  = add $sectionCount 1}}
                     {{ if gt $sectionCount $maxVisibleSections }}
-                      <a class="dropdown-item" href="{{ site.BaseURL | relLangURL }}#{{ partial "helpers/get-section-id.html" . }}">{{ .section.name }}</a>
+                      <a class="dropdown-item" href="#{{ partial "helpers/get-section-id.html" . }}">{{ .section.name }}</a>
```
hossainemruz, line 105: Same thing here.

layouts/partials/navigators/navbar.html
```diff
@@ -88,7 +88,7 @@
               {{ $sectionCount  = add $sectionCount 1}}
               {{ if le $sectionCount $maxVisibleSections }}
                 <li class="nav-item">
-                  <a class="nav-link" href="{{ site.BaseURL | relLangURL }}#{{ partial "helpers/get-section-id.html" . }}">{{ .section.name }}</a>
+                  <a class="nav-link" href="#{{ partial "helpers/get-section-id.html" . }}">{{ .section.name }}</a>
```
hossainemruz, line 91: This won't work on new updated navbar. We now show same menus on home and post pages. So, we can't have relative link here.

hossainemruz (commented):
Thank you @daviddavo for the PR. Unfortunately, this change breaks the new updated navbar behavior.

daviddavo:
I completely forgot about the blog feature. I think the solution would be removing the base URL just when we are on the "main" page, and keeping it on all other pages.

hossainemruz:
>I completely forgot about the blog feature. I think the solution would be removing the base URL just when we are on the "main" page, and keeping it on all other pages.

Yes. That should work. You can use `.IsHome` to determine whether it is home page or not.
Here is an example:
https://github.com/hugo-toha/toha/blob/c943f39617c1f2b90f160df07c60fa4ec9d0be7d/layouts/partials/navigators/navbar.html#L69C63-L69C63

daviddavo:
Should be fixed! I ended up creating a helper to avoid repeating code.

hossainemruz (approved):
Updated to fix home section URL

PR #807, opened 2023-10-15
Fix featured post section after update

BernatBC:
### Issue
After major update of the theme, featured posts section was not rendered.

### Description
Changed from old "recent-post" card to new "post" card.

### Test Evidence
When having this section enabled, now it renders fine.


hossainemruz (approved):
Thank you @BernatBC

PR #815, opened 2023-10-17
Justify post content (issue #831: Text in post is not justified)

### LINKED ISSUE #813: Text in post is not justified
## Expected Behavior
Content text in post page should be justified.

## Current Behavior
Content text in post page is not justified.

## Possible Solution

Modifying css should be enough.

## Steps to Reproduce

1. Visit https://hugo-toha.github.io/posts/markdown-sample/ and scroll down.

### Screenshots

image

BernatBC:
### Issue

Fixes #813.

### Description

Had to justify `p` element by default cause in some cases the about section also had problems. Had to center the date displayed in posts. There's probably a more neat way to do it.

### Test Evidence

image

You can also visit [bernatbc.tk/posts/](https://bernatbc.tk/posts/) and select any post.

### DIFF
--- a/assets/styles/components/texts.scss
+++ b/assets/styles/components/texts.scss
@@ -82,6 +82,8 @@ html[data-theme='dark'] {
 
   p {
     color: get-dark-color('text-color');
+    text-align: justify;
+    text-justify: inter-word;
   }
 
   input {
--- a/assets/styles/components/texts.scss
+++ b/assets/styles/components/texts.scss
@@ -82,8 +82,6 @@ html[data-theme='dark'] {
 
   p {
     color: get-dark-color('text-color');
-    text-align: justify;
-    text-justify: inter-word;
   }
 
   input {
--- a/assets/styles/components/texts.scss
+++ b/assets/styles/components/texts.scss
@@ -12,6 +12,8 @@ strong {
 
 p {
   color: get-light-color('text-color');
+  text-align: justify;
+  text-justify: inter-word;
 }
 
 input {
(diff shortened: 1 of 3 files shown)


assets/styles/components/texts.scss
```diff
@@ -80,6 +82,8 @@ html[data-theme='dark'] {
 
   p {
     color: get-dark-color('text-color');
+    text-align: justify;
+    text-justify: inter-word;
```
hossainemruz, line 86: You don't need to repeat these lines. Only add CSS that should be changed for dark mode.

layouts/_default/single.html
```diff
@@ -39,7 +39,7 @@
         <div class="author-profile ml-auto align-self-lg-center">
           <img class="rounded-circle" src='{{ partial "helpers/get-author-image.html" . }}' alt="Author Image">
           <h5 class="author-name">{{ partial "helpers/get-author-name.html" . }}</h5>
-          <p class="text-muted">{{ .Page.Date | time.Format ":date_full" }}{{ if site.Params.features.readingTime }} | {{ .ReadingTime }} minute{{ if (ne .ReadingTime 1) }}s{{ end }}{{ end }}</p>
+          <p class="text-muted" style="text-align: center;">{{ .Page.Date | time.Format ":date_full" }}{{ if site.Params.features.readingTime }} | {{ .ReadingTime }} minute{{ if (ne .ReadingTime 1) }}s{{ end }}{{ end }}</p>
```
hossainemruz, line 42: Please don't use inline css. Just add a `p` element under `author-profile` and apply CSS there:
https://github.com/hugo-toha/toha/blob/5b31653ed24d1607edbfe0bb598a381b72bf4deb/assets/styles/layouts/single.scss#L42

hossainemruz (changes requested):
Thank you for the PR @BernatBC. Left few comments.

assets/styles/components/texts.scss
BernatBC: Ok, thank you for this advice. Now it's fixed.

layouts/_default/single.html
BernatBC: Ok, now I added the text-align into the css file.

hossainemruz (approved):
LGTM 🚀

PR #821, opened 2023-10-23
Fixed localization of post cards

Kyserbyte:
### Issue
The post card component in the post list is not well localized

### Description
Now the post card id rightly localized in all possible languages supported by the theme.
The date and the time to read formatting are impacted by the changes
I also added some new translations for the Italian language


i18n/en.toml
```diff
@@ -124,3 +124,9 @@ other = "Liability Notice"
 
 [search]
 other = "Search"
+
+[minute]
+other = "minute"
+
+[minutes]
+other = "minutes"
```
hossainemruz, line 132: We don't need to define two variables for minute. You can just do:
```
[minute]
one = "minute"
other = "minutes"
```
This will automatically use `minute` where singular is used and `minutes` where plural is used.

layouts/partials/cards/post.html
```diff
@@ -12,12 +12,18 @@ <h5 class="card-title">{{ .Title }}</h5>
...
       <div class="card-footer">
-        <span class="float-left">{{ .Date.Format "January 2, 2006" }}{{ if site.Params.features.readingTime }} | {{ .ReadingTime }} minute{{ if (ne .ReadingTime 1) }}s{{ end }}{{ end }}</span>
+        <span class="float-left">
+            {{ .Date | time.Format ":date_full" }}
+            {{ if site.Params.features.readingTime }} | {{ .ReadingTime }} 
+              {{ if (ne .ReadingTime 1) }}
+                {{ i18n "minutes" }}
+              {{ else }}
+                {{ i18n "minute" }}
+              {{ end }}
```
hossainemruz, line 22: We don't need to check for singular or plural. We can just pass the count to `i18n` function and it will automatically chose singular or plural based on that count. For example:
```
{{ i18n "minute" .ReadingTime }}
```

hossainemruz (changes requested):
Thank you @Kyserbyte for the PR. I have requested a change. Please, take a look.

hossainemruz:
@Kyserbyte I have updated the PR. Needed a small change.

hossainemruz (approved):

PR #850, opened 2023-12-19
Fixed and improved circular progess animation on about section

Kyserbyte:
### Description
These changes was about to fix some behaviors about the circular progress animations.
Currently the circular progress only supports values greater then 50% and it breaks if you put 100%.
Please look at the "Problem Solving" and "Comunication" soft skills in following screenshots that should be 100 and 40 but renders as 50.

Screenshot 2023-12-19 alle 11 01 01
Screenshot 2023-12-19 alle 11 01 13

### Test Evidence
After the fix, now it is possible to use all values between 0 - 100 with steps (or granularity) of 5.
Look at the screenshot that render the same data of previous one.

Screenshot 2023-12-19 alle 11 01 50

### DIFF
--- a/assets/styles/sections/about.scss
+++ b/assets/styles/sections/about.scss
@@ -20,20 +20,25 @@ $progress-bar-colors: (
 }
 
 @mixin circular-progress-animation-breakpoints() {
-  $progress: 50;
+  $progress: 0;
   $duration: 0;
+  $delay: 1.8;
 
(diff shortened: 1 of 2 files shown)


hossainemruz:
Thank you @Kyserbyte for the PR. I will test and merge soon.

hossainemruz (approved):

PR #854, opened 2023-12-21
Specify European Portuguese

SrNetoChan:
### Issue

I did not report an issue, but with the brasilian portuguese available it seems that sites with the pt language now use pt-br, instead of pt

### Description

Updated some translations and renamed the file from pt to pt-pt


hossainemruz (approved):
Thank you @SrNetoChan

PR #866, opened 2024-01-19
Add select maximum number of posts per page

BernatBC:
### Issue
None, it's just a feature.

### Description
With this parameter you can select the maximum number of posts that will appear to each page. It will default to 12 (the value that we were using before). To configure it, just add the following lines under the `params` section of `config.yaml`:
```yaml
postsPage:
    maxPostsPerPage: 20 # Default: 12
```

Before I add this into the guides and the example file, tell me if you want to change the name of the parameter, or if I have to change the location of it.

#### maxPostsPerPage=5
image
#### maxPostsPerPage=10
image

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -32,7 +32,11 @@
   <div class="content container-fluid" id="content">
     <div class="container-fluid post-card-holder" id="post-card-holder">
       {{ $posts := where .RegularPagesRecursive "Layout" "!=" "search" }}
-      {{ $paginator := .Paginate $posts 12 }}
+      {{ $numShow := 12}}
+      {{ if site.Params.postsPage.maxPostsPerPage}}
+        {{ $numShow = site.Params.postsPage.maxPostsPerPage }}
+      {{ end }}
+      {{ $paginator := .Paginate $posts $numShow }}
       {{ range $paginator.Pages }}
         {{ if .Layout }}
           {{/* ignore the search.md file*/}}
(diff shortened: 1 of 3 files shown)


layouts/_default/list.html
```diff
@@ -32,7 +32,11 @@
   <div class="content container-fluid" id="content">
     <div class="container-fluid post-card-holder" id="post-card-holder">
       {{ $posts := where .RegularPagesRecursive "Layout" "!=" "search" }}
-      {{ $paginator := .Paginate $posts 12 }}
+      {{ $numShow := 12}}
+      {{ if site.Params.postsPage.maxPostsPerPage}}
```
hossainemruz, line 36: Let's call the settings `pagination` and put it under `features`. So, the configuration will look like:
```yaml
params:
  features:
    pagination:
       maxPostsPerPage: 10
```

hossainemruz: You can use [default](https://gohugo.io/functions/compare/default/) function to simplify the code little bit. For example:

```
{{ $numShow = site.Params.features.pagination.maxPostsPerPage | default 12}}
```

hossainemruz (changes requested):
Thank you @BernatBC for the PR. Look like a reasonable feature to add. I just requested two small adjustments. Otherwise, the PR looks good.

hossainemruz (approved):
Thank you @BernatBC!

hossainemruz (changes requested):
Look like the build is failing. Can you please fix that?

BernatBC:
I added it to the example site.
If you want, I can add it to the guides aswell.

BernatBC:
I'll fix it now

hossainemruz (approved):

PR #890, opened 2024-02-11
Fix Tags sidebar font color

BernatBC:
### Issue

Tags sidebar has wrong font color

### Description
Sets the same css as normal sidebar

### Test Evidence

Before:
image
image

After:
image
image


hossainemruz (approved):
Thank you @BernatBC

PR #891, opened 2024-02-11
Fix anchors in post cards

### LINKED ISSUE #868: Bug: extra anchor in post card tags
## Expected Behavior
Tags in post cards shouldn't have the extra anchor.

## Current Behavior
Tags in post cards have an extra anchor, which will result to having an extra button in the tags.

## Steps to Reproduce

1. Visit https://hugo-toha.github.io/
2. Scroll down to Recent Posts, and you'll see this in the Rich Content post

### Screenshots
bug

## Detailed Description

I've been investigating for a while, and notticed that the whole card anchor adds the anchor to the list, couldn't find an easy way to fix it.

BernatBC:
### Issue

Fixes #868 

### Description
Divided anchor between hero image and post card content so the anchor won't be inherited to the tags. It's not the neatest solution, but works.

### Test Evidence
image


hossainemruz (approved):
Thank you @BernatBC

PR #895, opened 2024-02-14
Add tags in post cards from search results

BernatBC:
### Issue
When tags are enabled in post cards, they don't appear in the post search results.

### Description
As you can see from the screenshot, if posts have different tags, the height is not adjusted at all. I've been looking how to get it fixed for a while, but I couldn't find the solution. If someone finds it, please tell me.

The other issue I have, is that at the moment I can only show tags when `enable: true`. The tags should be toggled when `on_card: true`.

#### Tags enabled:
image

#### Tags disabled:
image

### DIFF
--- a/assets/scripts/pages/search.js
+++ b/assets/scripts/pages/search.js
@@ -1,5 +1,6 @@
 import Fuse from 'fuse.js'
 import Mark from 'mark.js'
+import * as params from '@params'
 
 window.addEventListener('DOMContentLoaded', () => {
   const summaryInclude = 60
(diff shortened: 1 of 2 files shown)


BernatBC:
After fixing both issues, now it's ready:
image

assets/scripts/pages/search.js
```diff
@@ -79,14 +79,21 @@ window.addEventListener('DOMContentLoaded', () => {
       // pull template from hugo template definition
       const templateDefinition = document.getElementById('search-result-template').innerHTML
       // replace values
+      function adaptTags() {
+        const tags = value.item.tags;
```
hossainemruz, line 83: You need to check if tags null or not. The search on deploy preview fails with this error:
```bash
application.e4d15ee8d45ef1141f3853aff6a4dcb0489644435baa1e71cf89f5e1c292ade8.js:279 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'forEach')
    at y (application.e4d15ee8d45ef1141f3853aff6a4dcb0489644435baa1e71cf89f5e1c292ade8.js:279:32904)
    at application.e4d15ee8d45ef1141f3853aff6a4dcb0489644435baa1e71cf89f5e1c292ade8.js:279:33145
    at Array.forEach (<anonymous>)
    at n (application.e4d15ee8d45ef1141f3853aff6a4dcb0489644435baa1e71cf89f5e1c292ade8.js:279:32361)
    at application.e4d15ee8d45ef1141f3853aff6a4dcb0489644435baa1e71cf89f5e1c292ade8.js:279:32212
```

hossainemruz:
Look like some CSS was not applied properly.

This is what search result look like on your PR:
image

This is what it is look like in the demo site.

image

PR #914, opened 2024-03-27
add hebrew translation

skamensky:
### Description

Add hebrew translation


hossainemruz (commented):
Thank you @skamensky! Can you please, add Hebrew as `Available Translations` in the README file? Otherwise, looks good.

PR #939, opened 2024-05-16
Turkish Localization Updated

kemalkolcuoglu:
Turkish localization updated.


hossainemruz (approved):
Thank you @kemalkolcuoglu

PR #951, opened 2024-06-17
Allow for personalized titles for Blog and Note pages

maiani:
### Description

Simple change to allow personalized titles for Blog and Notes pages.
At the moment, this affects only the top navbar and not the sidebars.


hossainemruz (approved):
Thank you @maiani

PR #963, opened 2024-08-31
Add copy code button

### LINKED ISSUE #954: Feature Request - Add Copy Button to Code Blocks

It would be interesting to add the "copy code" button in the code blocks in the blog. Just similar to ChatGPT or Gemini

### Describe alternatives you've considered

I am not a professional web-developer, so If found how to do it here:
 https://www.dannyguo.com/blog/how-to-add-copy-to-clipboard-buttons-to-code-blocks-in-hugo/

BernatBC:
### Issue
- Fixes #954

### Description
Adds a copy code button in all code blocks. To enable this feature, you'll need to add the following lines into `config.yaml`/`hugo.yaml`:
```yaml
params:
  features:
    # Enables copy code button
    copyCodeButton:
      enable: true
```

I will add this into [hugo-toha/hugo-toha.github.io](https://github.com/hugo-toha/hugo-toha.github.io) and  [hugo-toha/guides](https://github.com/hugo-toha/guides) 
### Test Evidence

Screenshot_20240831_112156


hossainemruz:
Hi @BernatBC! I will review on the weekend. Giving a quick look on the deploy preview, I noticed these:

1. The copy button looks large. We have to reduce its size little bit.
2. Need to adjust margin/padding so that the copy icon is at the same level as the first line.
2. It does not have any feedback on hover.
3. It does not have any feedback when clicked.

You can follow the GitHub style. Also, we need to think about where to put the configuration. It will be better to unify all markdown related configurations under a single section.

BernatBC:
Hi @hossainemruz,
I've made some tweaks to it. Regarding your points:
1. I've reduced a little bit the icon size.
2. I've readjusted a little bit the margin, so now it's aligned properly.
3. When on hover, the icon has the same behavior as any other link. I have also added a mouse tooltip. Should we translate the string "copy"?
4. I've been thinking about it, and I'm not sure what's the best feedback to give after pressing the button. Maybe a little text "Copied!"? If that's the case, will we have to add it to localization as well?

hossainemruz:
> Hi @hossainemruz, I've made some tweaks to it. Regarding your points:
> 
> 1. I've reduced a little bit the icon size.
> 2. I've readjusted a little bit the margin, so now it's aligned properly.
> 3. When on hover, the icon has the same behavior as any other link. I have also added a mouse tooltip. Should we translate the string "copy"?
> 4. I've been thinking about it, and I'm not sure what's the best feedback to give after pressing the button. Maybe a little text "Copied!"? If that's the case, will we have to add it to localization as well?

Awesome. Thank you. I will take a look soon.

BernatBC:
Hi @hossainemruz,
it looks great!
Thank you once again

hossainemruz:
@BernatBC I was wondering if we should combine these kind of feature configurations under common section? That will make them easily discover-able. For example:

```yaml
params:
  features:
    utility: # not sure what should be the name. feel free to suggest
      copyCodeButton:
         enable: true # add a button to copy code on a code blocks. default "true"
      readingTime:
         enable: true # show estimated reading time on reading page. default: "true"
``` 
Maybe few other settings can go there too.

BernatBC:
Looks good to me, but will need to be careful with this change, as it will introduce breaking changes for "readingTime"

hossainemruz:
> Looks good to me, but will need to be careful with this change, as it will introduce breaking changes for "readingTime"

Keeping as it is now. If we want to group them under utility section, the configuration will be:
```yaml
params:
  features:
    utility:
      enabled: true
      services:
        copyCodeButton:
          enable: true # add a button to copy code on a code blocks. default "true"
        readingTime:
          enable: true # show estimated reading time on reading page. default: "true"
```
The utility functions does not have anything in common. So, it does not make sense to put them under same service.

hossainemruz (approved):

rageSpin:
Thanks 💯

Issue #955, opened 2024-07-03
bug Google Analytics has stopped working


javierasping:
Since June 3rd, when the Git bot applied the following commit --> https://github.com/javierasping/sentinel/commit/0cf1c666d6f4bbddc971541cbe24c068b3ab6c9f, my Google Analytics stats have stopped working. I have been looking into it and the script is not being inserted into the HTML code. My Google tag is G-GVDYVWJLRH.

When I search for it in the generated HTML code, I cannot find it.

```bash
javiercruces@HPOMEN15:~/Documentos/sentinel/public$ grep -r "G-GVDYVWJLRH" .
```

My website is in www.javiercd.es . 

I have tried to do as in previous cases where you helped me update to the latest version of Hugo as well as update the dependencies using the following commands:

```bash
# update to latest version of all modules
hugo mod get -u

# update the npm dependencies
hugo mod npm pack

# cleanup go.sum file
hugo mod tidy

# install npm dependencies
npm install
```
My current version of Hugo is:

```bash
javiercruces@HPOMEN15:~/Documentos/sentinel$ hugo version
hugo v0.128.1-0ff542b4b9cc7b5cb425bbab8bdb5aace81d0c03+extended linux/amd64 BuildDate=2024-07-02T06:46:41Z VendorInfo=gohugoio
```

Thank you very much in advance for your help, best regards!!

mgnkir:
Same problem here

deining:
> Since June 3rd, when the Git bot applied the following commit --> [javierasping/sentinel@0cf1c66](https://github.com/javierasping/sentinel/commit/0cf1c666d6f4bbddc971541cbe24c068b3ab6c9f), my Google Analytics stats have stopped working.

In your `config.yaml`, I see this [code](https://github.com/javierasping/sentinel/blob/2011ec6cf54f72bc6fb90238ba7325b57c66a100/config.yaml#L142C1-L148C27) which is incorrect IMHO:

```
# Enable Analytics
analytics:
  enabled: true
  services:
    # Google Analytics
    google:
      id: G-GVDYVWJLRH
```

According to the Hugo [documentation](https://gohugo.io/methods/site/config/#services), the Google analytics ID should be go here:

```
services:
  googleAnalytics:
    id: "G-GVDYVWJLRH"
```

In your `config.yaml`, I can't see a `services` section at the root level at all. Can you please add one and report back if that cures your problem?

[...]

PR #989, opened 2024-10-01
Fix google analytics feature

joyao:
### Issue
Fixes https://github.com/hugo-toha/toha/issues/955

### Description

Fix google analytics feature based on official document.
https://gohugo.io/templates/embedded/#google-analytics

### Test Evidence
image

### DIFF
--- a/layouts/partials/analytics.html
+++ b/layouts/partials/analytics.html
@@ -4,9 +4,7 @@
         {{ with .services }}
             <!-- Google Analytics -->
             {{ with .google }}
-                {{ $privacyConfig:= dict (slice "Site" "Config" "Privacy" "GoogleAnalytics") $.Site.Config.Privacy.GoogleAnalytics }}
-                {{ $analyticsConfig := dict (slice "Site" "Config" "Services" "GoogleAnalytics" "ID") .id }}
-                {{ template "_internal/google_analytics.html" (merge $privacyConfig $analyticsConfig) }}
+                {{ partial "google_analytics.html" . }}
             {{ end }}
 
             <!-- Counter.dev -->
--- a/layouts/partials/google_analytics.html
+++ b/layouts/partials/google_analytics.html
@@ -1,0 +1,22 @@
+{{ if site.Params.features.analytics.enabled }}
+  {{- with site.Params.features.analytics.services.google.id }}
+    {{- if strings.HasPrefix (lower .) "ua-" }}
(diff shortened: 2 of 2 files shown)


hossainemruz (commented):
Hi @joyao ! Thank you for the PR. Look like we are breaking compatibility here. With this change, following configuration will not work anymore:
```yaml
services:
  googleAnalytics:
    ID: G-MEASUREMENT_ID
```

Do we want to force user to use analytics configuration from `site.Params.analytics.services.google.id` ? In that case, we should also provide a way do specify `respectDoNotTrack` under `site.Params.analytics.services.google` section.

joyao:
> Hi @joyao ! Thank you for the PR. Look like we are breaking compatibility here. With this change, following configuration will not work anymore:
> 
> ```yaml
> services:
>   googleAnalytics:
>     ID: G-MEASUREMENT_ID
> ```
> 
> Do we want to force user to use analytics configuration from `site.Params.analytics.services.google.id` ? In that case, we should also provide a way do specify `respectDoNotTrack` under `site.Params.analytics.services.google` section.

I think both methods should be supported. I’ll fix it and thanks for your suggestion.

jamesdeluk:
So what's the current workaround for getting analytics working, manually having these two files within layouts?

joyao:
> So what's the current workaround for getting analytics working, manually having these two files within layouts?

Hi @jamesdeluk, this PR hasn't been merged yet. If you want to have this fixed feature, you can merge my branch or fix these two files manually in your project.

joyao:
> 
> Thanks. And which is the best `yaml` config of the two above?

This one:

``` yaml
analytics:
  enabled: true
  services:
    # Google Analytics
    google:
      id: G-H4LBG7NDFZ
      respectDoNotTrack: false
```

You don't need to make many changes to this, this section is already in the current hugo.yaml file. You just need to change the id.

jamesdeluk:
> This one:
> 
> ```yaml
> analytics:
>   enabled: true
>   services:
>     # Google Analytics
>     google:
>       id: G-H4LBG7NDFZ
>       respectDoNotTrack: false
> ```
> 
> You don't need to make many changes to this, this section is already in the current hugo.yaml file. You just need to change the id.

Great, thanks! And I guess `respectDoNotTrack` can be set to `true` and it'll still work?

joyao:
> 
> Great, thanks! And I guess `respectDoNotTrack` can be set to `true` and it'll still work?

Yes, it should work. :)

joyao:
Hi @hossainemruz, I believe it works as expected. Please take your time reviewing it. Thank you.

hossainemruz (approved):

Issue #1000, opened 2024-10-15
Feature Request: support for canonical links


davidgs:
### Describe the solution you'd like

Support for canonical links in front matter. I often post things to commercial sites, but *also* want to post them to my personal site later. In order for google and other search engines not to refuse to index those pages, I need to have a 
` <link rel="canonical" href="link-to-original-content />`
on my post. 

### Describe alternatives you've considered

This is the simplest solution. I have been using it for years on earlier versions of torah, so I thought it might be time to contribute it back.

davidgs:
see PR #1001

PR #1001, opened 2024-10-15
Add support for canonical links

davidgs:
### Issue
Fixes https://github.com/hugo-toha/toha/issues/1000

### Description

Add support for canonical links in frontmatter. If you post to an external site, and *also* want to post to your own site, you **must** use a `<link rel="canonical" ` in the header or search engines will refuse to index it (on either site!)

### Test Evidence
canonical1

canonical2

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -8,9 +8,7 @@
     {{ template "_internal/twitter_cards.html" . }}
     <!------ ADD PAGE SPECIFIC HEADERS ------->
     {{ block "header" . }} {{ end }}
-    {{ with .Params.relcanonical }}
-      {{ partial "misc/canonical.html" . }}
-    {{ end }}
+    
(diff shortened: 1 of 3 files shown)


hossainemruz:
HI @davidgs! Thank you for the PR. If I understand it correctly, this only necessary for post pages right? In that case, I will recommend to add it here:
https://github.com/hugo-toha/toha/blob/bc67ca11873afe2165e6b08cf38dd5e3a3fa735f/layouts/_default/single.html#L2

There is no need to separate this single line into a partial template.

davidgs:
As long as that works, I'm happy! Just having `real canonical` working is a win!

davidgs:
Works just fine. Moved to `single.html` and removed `canonical.html` and the code in `based.html`

hossainemruz (approved):
Thank you @davidgs

PR #1005, opened 2024-10-21
Fix input focus and placeholder color for footer section

joyao:
### Issue
Solved https://github.com/hugo-toha/hugo-toha.github.io/issues/348

### Description

Input text cannot be seen in dark mode. Fix it by modifying the css.

#### Before
image

#### After
Placeholder
image
Focus
image


hossainemruz (approved):
Thank you @joyao

PR #1017, opened 2024-12-13
 Adapt publication card to handle missing URLs

maiani:
### Description
This pull request enhances the `publication` card to improve its handling of missing URLs for authors and publications. 

Key changes include:
- **Conditional Link Handling**: The `href` parameter is completely removed from the `a` tags when no URL is available.  
- **Style Update**: Underlines are applied only to elements that contain active links.  
- **Button Visibility**: The `Details` button is hidden when no URL is provided for the paper.  

These updates ensure a cleaner, more intuitive user interface and prevent confusion caused by non-functional links or buttons.

Sorry for the duplicate pull request.


hossainemruz (approved):
Thank you @maiani

PR #1032, opened 2025-02-06
fix: Create text-justify css class

remyj38:
### Description

.text-justify is used on some texts but never defined


hossainemruz (approved):
It comes from bootstrap. Anyway, having this class defined here makes it little more independent of bootstrap. Thank you @remyj38 .

Issue #1053, opened 2025-06-01
Google analytics support seems to be broken


tduchateau:
Hi!

Support for analytics seems to be broken for me.

By following documentation and looking at various issues, I've tested several configurations, but with no success.

Test 1, in my `hugo.yaml` file: 
````
params:
  features:
    analytics:
      enable: true
      services:
        google:
          id: G-GBN86LC9K4
````

Test 2, in my `hugo.yaml` file: 
````
services:
  googleAnalytics:
    id: "G-GBN86LC9K4"

params:
  features:
    analytics:
      enable: false
````

But I can't see any partial being included with the googletagmanager stuff.

Can anyone help me?

Hugo version: 0.147.2
Toha theme: 4.9.0

I'd like to take this opportunity to thank you for all your work!

PR #1079, opened 2025-09-29
Implement 'enable' parameter support in GA settings

joyao:
### Issue
Fixes https://github.com/hugo-toha/toha/issues/1053

### Description

This change refactors the Google Analytics setting support to correctly handle both enabled (old) and enable (new) as valid parameter values.
This ensures backward compatibility while aligning with newer configurations. The code has also been refactored for better readability.

### Test Evidence
None


hossainemruz (approved):
Thank you @joyao

Issue #1126, opened 2026-01-15
Feature Request: No analytics on localhost


davidgs:
### Describe the solution you'd like

A way to disable loading analytics when running on `localhost` e.g. when running `hugo server -w`

### Describe alternatives you've considered

Right now the only way to do this is to comment out analytics in `theme.toml` when doing development, which is inelegant and you have to remember to turn them back on for deployment.

PR #1127, opened 2026-01-15
Disables analytics on localhost

davidgs:
### Issue
#1126

### Description

Prevents analytics services from loading when the site is accessed via localhost. This change ensures that development environments do not generate unnecessary or misleading analytics data.

### Test Evidence


hossainemruz (commented):
Thank you @davidgs for the PR. Added a comment. Please, take a look.

PR #1180, opened 2026-05-26
Fix publication author color in dark mode

maiani:
Fixes publication-card author names without URLs rendering with the wrong color in dark mode by explicitly applying theme text colors to the authors block.


hossainemruz (approved):
Thank you @maiani

PR #1195, opened 2026-06-23
fix: support external URLs and page bundle images in project cards

ru-admin:
## Summary

Project card template failed when `image` or `logo` pointed to:
1. **External URLs** (`https://...`) — `resources.Get` tried to resolve them as local filesystem paths, causing a build error.
2. **Page bundle resources** (e.g. `/posts/.../images/forest.jpg`) — `resources.Get` only resolves files in `assets/`, so the image was silently skipped.

## Changes

`layouts/partials/cards/project.html`:
- Added `hasPrefix` check for `http://` / `https://` — external URLs are rendered directly as `` without `resources.Get`.
- Added fallback: if `resources.Get` returns `nil` for a local path, render the path directly (supports page bundle images).
- Same logic applied to both `image` and `logo` fields.
- Add pic with external URL to exampleSite

## Backward Compatibility

Existing asset-based images continue to work through Hugo Pipes (Fit, SVG detection). No breaking changes to data schema.

### DIFF
--- a/exampleSite/data/en/sections/projects.yaml
+++ b/exampleSite/data/en/sections/projects.yaml
@@ -48,6 +48,7 @@ projects:
 
 - name: Rich Content
   url: "/posts/category/sub-category/rich-content/"
+  image: "https://toha-preview.hugo-themes.com/posts/category/sub-category/rich-content/images/forest.jpg"
   # Optional: control how project links open. Allowed values: "blank" (default) or "self".
   # "blank" opens links in a new tab, "self" opens them in the current tab.
   target: self
--- a/layouts/partials/cards/project.html
+++ b/layouts/partials/cards/project.html
@@ -8,6 +8,9 @@
         .repo .url }}target="{{ $targetAttr }}" rel="noopener" {{ end }}>
         {{ if .image }}
         <div class="card-head">
+          {{ if or (hasPrefix .image "http://") (hasPrefix .image "https://") }}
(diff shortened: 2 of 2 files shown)


hossainemruz:
This fallback fixes the `resources.Get` failure case, but it now renders local paths directly:

```html
<img class="card-img-top" src="{{ .image }}" ...>
<img class="card-img-xs" src="{{ .logo }}" ...>
```
That introduces a baseURL/subpath bug for the page-bundle case this PR is trying to support. For example, if a user configures:
```yaml
image: /posts/category/sub-category/rich-content/images/forest.jpg
```
and hosts the site under a subpath, such as GitHub Pages with:
```yaml
baseURL: https://example.github.io/my-site/
```
the rendered HTML will contain:
```yaml
src="/posts/category/sub-category/rich-content/images/forest.jpg"
```
which points to the domain root and will 404. Hugo should rewrite local URLs relative to the configured base path, e.g. `/my-site/posts/....`

The existing asset-resource path uses `RelPermalink`, so it already gets Hugo URL handling. The new non-asset local fallback should do the same kind of URL rewriting. External URLs should still be rendered directly, but local fallback paths should probably be piped through `relURL`, for example:
```html
<img class="card-img-top" src="{{ .image | relURL }}" alt="{{ .name }}" loading="lazy" />
```
and similarly for `.logo`:
```html
<img class="card-img-xs" src="{{ .logo | relURL }}" alt="{{ .name }}" />
```

PR #1204, opened 2026-07-29
Fix conditional for enabling analytics

### LINKED ISSUE #1178: Umami analytics not working
I've noticed the past months that analytics on my webpage are not being captured. After investigating more, I noticed the Umami script tag is no longer being added to the output HTML.

I browsed through the changes and reverting [this commit](https://github.com/hugo-themes/toha/pull/1127/changes) fixes the issue for me.

[Here is my website source code](https://git.korhonen.cc/FunctionalHacker/korhonen.cc) if you want to verify it's not something in my config.

I suspect this also affects other analytics platforms but I only have Umami to test with

JokaJukka:
### Issue

Fixes https://github.com/hugo-themes/toha/issues/1178

### Description
Fix nested `enabled` conditional in analytics params.

### Test Evidence
The bugged variant forced you to configure params as:
```
params:
  features:
    analytics:
      enabled: true
      services:
        enabled: true
        ...
```

Instead of the intended
```
params:
  features:
    analytics:
      enabled: true
      services:
        ...
```

### DIFF
--- a/layouts/partials/analytics.html
+++ b/layouts/partials/analytics.html
@@ -1,8 +1,7 @@
 <!-- Add Analytics if enabled in configuration -->
 {{ with site.Params.features.analytics }}
-    {{ if or .enable .enabled }}
+    {{ if and hugo.IsProduction (or .enable .enabled) }}
         {{ with .services }}
-        {{ if and hugo.IsProduction (or .enable .enabled)}}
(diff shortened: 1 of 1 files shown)


hossainemruz (approved):
Thank you @JokaJukka

Issue #1208, opened 2026-08-06
Internal links (customMenus, project cards) don't respect baseURL when it has a subpath


yinchi:
## Expected Behavior
Internal links (e.g. a `customMenus` entry pointing to an internal page, or a
project card's `.url` pointing to an internal post) should resolve correctly
under any configured `baseURL`, including one with a subpath — e.g. a GitHub
Pages project site like `https://user.github.io/repo/`.

## Current Behavior
These links are rendered as a raw, unprocessed `href="{{ .url }}"` with no
baseURL handling, so they only produce a correct path when `baseURL` has no
subpath. On a subpath deployment the link is missing the subpath entirely and
points to the wrong location — e.g. `/posts` instead of `/repo/posts`.

Two confirmed instances:
- `customMenus` entries (`navbar.html`, `footer.html`)
- Project card links/buttons (`cards/project.html`, `.url`) — the theme's own
  `exampleSite` already demonstrates an internal-path use case for this field
  (the "Rich Content" project in `data/en/sections/projects.yaml`), so this
  isn't hypothetical.

## Possible Solution
Distinguish internal (no `http://`/`https://` prefix) from external URLs, and
resolve internal ones by joining them onto `site.BaseURL | relLangURL` (the
same pattern already used ad hoc for the blog/notes nav links).

## Steps to Reproduce
1. In `exampleSite`, note the "Rich Content" project card in
   `data/en/sections/projects.yaml`, which uses an internal `url`
   (`/posts/category/sub-category/rich-content/`).
2. Build with a subpath baseURL:
   `hugo --gc --minify --baseURL "https://example.com/subpath/"`
3. Inspect the rendered `href` for that project card.
4. Observe it renders as `/posts/category/sub-category/rich-content` — missing
   the `/subpath` prefix — while other internal links on the same page (e.g.
   section anchors, the post's own card elsewhere via `.RelPermalink`) are
   correctly prefixed with `/subpath`.

### Screenshots
N/A — this is a server-rendered HTML output issue (wrong `href` value), not a
visual/layout bug.

### Browsers Affected
N/A — the bug is in Hugo's build-time template rendering, not
browser-specific.

## Detailed Description
Several `href` attributes in the theme are set directly from a data field
(`{{ .url }}`) without passing through any Hugo URL-resolution function
(`relLangURL`, `absLangURL`, `RelPermalink`, etc.). This only happens to
produce a correct URL when `baseURL` has no subpath — on a project site or any
[...]

PR #1209, opened 2026-08-06
fix: resolve internal links against baseURL for non-root deployments (#1208)

yinchi:
### Issue

#1208 

### Description

Several internal links (e.g. in navbar.html, footer.html, cards/project.html) used plain {{ .url }} without baseURL handling. This only worked when baseURL had no subpath; it breaks on any non-root deployment, e.g. some site served from "person.dev/me/" rather than the domain root "person.dev".

The problematic call sites now distinguish external (http/https) URLs (left untouched) from internal ones, which are now resolved via a new `helpers/get-page-url.html` partial. `helpers/get-section-url.html` is refactored to use the same helper instead of inlining `site.BaseURL | relLangURL` directly. The brand, home, blog, and notes links in navbar.html, which already inlined this same `path.Join (site.BaseURL | relLangURL) ...` pattern ad hoc, are likewise refactored onto the shared helper, so there is now a single place that resolves internal links against baseURL.

### Test Evidence

Reproduced against this repo's own `exampleSite`, built with a non-root `baseURL`
to simulate a subpath deployment (e.g. a GitHub Pages project site):

```sh
cd exampleSite
hugo --gc --minify --cleanDestinationDir --baseURL "http://localhost:1313/subpath-test/"
```

**Before the fix**, internal links in the affected spots ignored the subpath entirely:
- `customMenus` internal entry → rendered as `/posts` (missing `/subpath-test`)
- The "Rich Content" project card's `.url` (`/posts/category/sub-category/rich-content/`,
  Toha's own example data) → rendered as `/posts/category/sub-category/rich-content`
  (missing `/subpath-test`)

**After the fix**, both resolve correctly:
- `customMenus` internal entry → `/subpath-test/posts`
- "Rich Content" project card (both the card link and the "Details" button) →
  `/subpath-test/posts/category/sub-category/rich-content`

External URLs (e.g. the `Docs` customMenu entry, GitHub repo-linked project cards)
are confirmed unaffected — still rendered as-is.

The `navbar.html` brand/home/blog/notes links, which already inlined the correct
`path.Join (site.BaseURL | relLangURL) ...` pattern before this PR, were refactored
onto the new shared helper and re-verified to produce identical output
(`/subpath-test`, `/subpath-test#home`, `/subpath-test/posts`, `/subpath-test/notes`)
— confirming the refactor didn't regress already-working links.

`npm run lint` passes with no new violations.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -86,11 +86,7 @@
             {{ if $customMenusEnabled }}
               {{ range $customMenus }}
                 {{ if .showOnFooter }}
-                    {{ $menuURL := .url }}
-                    {{/* Process internal links (not http[s]://) */}}
-                    {{ if not (or (hasPrefix $menuURL "http://") (hasPrefix $menuURL "https://")) }}
-                      {{ $menuURL = partial "helpers/get-page-url.html" $menuURL }}
-                    {{ end }}
+                    {{ $menuURL := partial "helpers/get-page-url.html" .url }}
                     <li class="nav-item">
                       <a class="smooth-scroll" href="{{ $menuURL }}">{{ .name }}</a>
                     </li>
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -86,8 +86,13 @@
             {{ if $customMenusEnabled }}
               {{ range $customMenus }}
                 {{ if .showOnFooter }}
+                    {{ $menuURL := .url }}
+                    {{/* Process internal links (not http[s]://) */}}
+                    {{ if not (or (hasPrefix $menuURL "http://") (hasPrefix $menuURL "https://")) }}
+                      {{ $menuURL = partial "helpers/get-page-url.html" $menuURL }}
+                    {{ end }}
                     <li class="nav-item">
-                      <a class="smooth-scroll" href="{{ .url }}">{{ .name }}</a>
+                      <a class="smooth-scroll" href="{{ $menuURL }}">{{ .name }}</a>
                     </li>
                 {{ end }}
               {{ end }}
--- a/layouts/partials/helpers/get-page-url.html
+++ b/layouts/partials/helpers/get-page-url.html
@@ -1,1 +1,8 @@
-{{- path.Join (site.BaseURL | relLangURL) (strings.Trim . "/") -}}
+{{/* Capture any trailing slash, dropped by path.Join, so we can add it back if needed. */}}
+{{- $hadTrailingSlash := strings.HasSuffix . "/" -}}
+{{- $resolved := path.Join (site.BaseURL | relLangURL) (strings.Trim . "/") -}}
+{{/* Add back any trailing slash if the original path had one, unless the resolved path is just "/" */}}
+{{- if and $hadTrailingSlash (ne $resolved "/") -}}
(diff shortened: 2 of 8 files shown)


layouts/partials/helpers/get-page-url.html
```diff
@@ -0,0 +1,8 @@
+{{/* Capture any trailing slash, dropped by path.Join, so we can add it back if needed. */}}
+{{- $hadTrailingSlash := strings.HasSuffix . "/" -}}
+{{- $resolved := path.Join (site.BaseURL | relLangURL) (strings.Trim . "/") -}}
+{{/* Add back any trailing slash if the original path had one, unless the resolved path is just "/" */}}
+{{- if and $hadTrailingSlash (ne $resolved "/") -}}
+  {{- $resolved = printf "%s/" $resolved -}}
+{{- end -}}
+{{- $resolved -}}
```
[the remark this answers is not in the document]
yinchi: Implemented the proposed change in [0498979](https://github.com/hugo-themes/toha/pull/1209/commits/0498979f10347cf8bc368372641d5a21ad54f2bb) and removed the per-call-site guards, which only checked for http/https.

layouts/partials/navigators/navbar.html
```diff
@@ -136,18 +136,19 @@
...
         {{ end }}
         {{ if $notesEnabled }}
           <li class="nav-item">
-            <a class="nav-link" id="note-link" href="{{ path.Join (site.BaseURL | relLangURL) "notes" }}">{{ $notesTitle }}</a>
+            <a class="nav-link" id="note-link" href="{{ partial "helpers/get-page-url.html" "notes" }}">{{ $notesTitle }}</a>
           </li>
         {{ end }}
         {{ range $customMenus }}
             {{ if (not .hideFromNavbar) }}
+              {{ $menuURL := partial "helpers/get-page-url.html" .url }}
```
hossainemruz, line 149: Minor edge case: the helper is called unconditionally, so a menu entry without `url` now resolves to the site root instead of rendering `href=""` as before (verified with a url-less test entry → `href=/subpath-test` on this branch). Consider guarding with `{{ with .url }}` or `{{ if .url }}`, mirroring the `if $projectURL` guard in `cards/project.html`. Not blocking.

layouts/partials/footer.html
```diff
@@ -86,8 +86,9 @@ <h5>{{ i18n "navigation" }}</h5>
             {{ if $customMenusEnabled }}
               {{ range $customMenus }}
                 {{ if .showOnFooter }}
+                    {{ $menuURL := partial "helpers/get-page-url.html" .url }}
```
hossainemruz, line 89: Same edge case as the navbar custom-menu note: a footer menu entry without `url` resolves to the site root rather than `""`. Suggest the same `if`/`with` guard for parity. Not blocking.

hossainemruz (commented):
Thank you @yinchi. Added few comments.

layouts/partials/footer.html
yinchi: I want to confirm the expected behavior: should it be `{{ $menuURL = "javascript:void(0)" }}` thus truly going nowhere for a `url`-less entry?  This matches a usage of `javascript:void(0)` I've found in `layouts/partials/cards/project.html`.

